PR #4, opened 2016-11-25
fixed cppflags

ltalirz:
Sorry that -fPIC got lost...
And CFLAGS should have been CPPFLAGS

Should now work on both Macos and Linux

### DIFF
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -1,6 +1,6 @@
 CC=g++
 LD=g++
-CPPFLAGS= -std=c++11 -O3 -ffast-math -ftree-vectorize 
+CPPFLAGS= -std=c++11 -O3 -ffast-math -ftree-vectorize -fPIC
 LDFLAGS= -shared
 
 all: PP GU MP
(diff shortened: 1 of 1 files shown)


ondrejkrejci:
Hi all Probe Particle Lords,

I'm at the moments in train, but once I'll be back on solid internet 
connection I'm planning to update XSF writing and reading procedure (and 
geometry reading of XSF files). The XSF files has the first and the last 
cubicle the same in every direction. This leads to small inconsistency 
of when the FFLJ and FFel is calculated. In the new procedure are the 
repeated cubicles added/removed to/from the written/readed XSF, so the 
FFel and FFLJ are on the same place, BUT now the nDim will be 
number_of_cubicles_in_the_XSF file -1. For those who uses geometry from 
xyz files and specifies gridN they should keep this in mind. I'll put 
this into documentation at 
http://nanosurf.fzu.cz/wiki/doku.php?id=probe_particle_model. Reading 
both - geometry and electrostatic - from XSF causes no problems.
I'm going to change the procedures on the master, dev and PP-STM 
branches. If somebody is against this change please let me know.
Kings regards,
Ondrej

Upozornění / Disclaimer <http://www.fzu.cz/e-mail-upozorneni-disclaimer>

PR #36, opened 2022-10-28
Initial version of sphinx docs

### LINKED ISSUE #34: Create initial version of sphinx docs

NikoOinonen:
Fixes #34 

This includes just the bare minimum needed for compiling the html doc pages. Requires sphinx, and sphinx_rtd_theme to be installed on the system (pip install sphinx sphinx_rtd_theme). I made ocl.AFMulator.py as an example of a module with correctly formatted doc-strings that get rendered properly on the html page.

To make the html docs, navigate to doc/sphinx and issue `make html`, and then open build/html/index.html to view the result.

Currently throws a lot of errors because some of the modules cannot be properly imported due to undefined variables etc., but most of the pages get generated successfully.


pyProbeParticle/ocl/AFMulator.py
```diff
@@ -36,7 +37,7 @@ class AFMulator():
         iZPPs: int. Element of probe particle.
         QZs: Array of length 4. Position tip charges along z-axis relative to probe-particle center in angstroms.
         Qs: Array of length 4. Values of tip charges in units of e. Some can be zero.
-        rho: Dict or MultipoleTipDensity. Tip charge density. Used with Hartree potentials. Overrides QZs and Qs.
+        rho: Dict or :class:`.MultipoleTipDensity`. Tip charge density. Used with Hartree potentials. Overrides QZs and Qs.
```
ondrejkrejci, line 40: I have seen, that everything is not affecting the running part of the code. These are only comment, I understand that adding the `:class:` is required by the automatic documentation, on the other hand it is pretty confusing for reader, I think. Is there some other way around?

pyProbeParticle/ocl/AFMulator.py
```diff
@@ -113,18 +114,21 @@ def __init__( self,
     def eval(self, xyzs, Zs, qs, pbc_lvec=None, rot=np.eye(3), rot_center=None, REAs=None, X=None ):
         '''
         Prepare and evaluate AFM image.
+
         Arguments:
             xyzs: np.ndarray of shape (num_atoms, 3). Positions of atoms in x, y, and z.
             Zs: np.ndarray of shape (num_atoms,). Elements of atoms.
-            qs: np.ndarray of shape (num_atoms,) or HartreePotential or None. Charges of atoms or hartree potential.
+            qs: np.ndarray of shape (num_atoms,) or :class:`.HartreePotential` or None. Charges of atoms or hartree potential.
```
ondrejkrejci, line 121: The same here?

pyProbeParticle/ocl/AFMulator.py
```diff
@@ -133,7 +137,7 @@ def eval(self, xyzs, Zs, qs, pbc_lvec=None, rot=np.eye(3), rot_center=None, REAs
     
     def __call__(self, xyzs, Zs, qs, pbc_lvec=None, rot=np.eye(3), rot_center=None, REAs=None, X=None):
         '''
-        Makes object callable. See eval for input arguments.
+        Makes object callable. See :meth:`eval` for input arguments.
```
ondrejkrejci, line 140: Here for the `:meth:`

pyProbeParticle/ocl/AFMulator.py
```diff
@@ -222,10 +226,11 @@ def setQs(self, Qs, QZs):
     def prepareFF(self, xyzs, Zs, qs, pbc_lvec=None, rot=np.eye(3), rot_center=None, REAs=None):
         '''
         Prepare molecule parameters and calculate force field.
+
         Arguments:
             xyzs: np.ndarray of shape (num_atoms, 3). Positions of atoms in x, y, and z.
             Zs: np.ndarray of shape (num_atoms,). Elements of atoms.
-            qs: np.ndarray of shape (num_atoms,) or HartreePotential or None. Charges of atoms or hartree potential.
+            qs: np.ndarray of shape (num_atoms,) or :class:`.HartreePotential` or None. Charges of atoms or hartree potential.
```
ondrejkrejci, line 233: And finally here for the `class`.

pyProbeParticle/ocl/AFMulator.py
```diff
@@ -422,7 +430,7 @@ def quick_afm(file_path, scan_size=(16, 16), offset=(0, 0), distance=8.0, scan_s
         sigma: float. Width of tip charge distribution.
         num_heights: int. Number of different heights to scan.
         amplitude: float. Oscillation amplitude in angstroms.
-        out_dir: str or None. Output folder path. If None, defaults to afm_ + input_file_name.
+        out_dir: str or None. Output folder path. If None, defaults to "./afm\_" + input_file_name.
```
ondrejkrejci, line 433: Why is there the backslach after "afm"?

Otherwise it is completely OK with me.

ondrejkrejci (commented):
My comments are probably minor, but if you just tell me, that there is no other way around, then I can Approve the merging. Anyway, thank you for working on this!

pyProbeParticle/ocl/AFMulator.py
NikoOinonen: These directives (`:class:`, etc) are not strictly necessary for the documentation, but they are just a convenience thing that adds a clickable link into the html page to take you to the definition of the class. The idea in the end is that you would not need to go to the source code to read these comments, but you would rather just go to the webpage where it looks nice once we get it up and running. Of course it's a bit more work and it's probably something we should discuss in a meeting if we want to have it. I would anyway merge it like this for now as an example of things that you can do.

pyProbeParticle/ocl/AFMulator.py
NikoOinonen: See above.

pyProbeParticle/ocl/AFMulator.py
NikoOinonen: See above.

pyProbeParticle/ocl/AFMulator.py
NikoOinonen: See above.

pyProbeParticle/ocl/AFMulator.py
NikoOinonen: The backslash is an escape character here. Without it, the 'afm' string is interpreted as a hyperlink on the html page.

pyProbeParticle/ocl/AFMulator.py
ondrejkrejci: Ok, I see, just asking. Good to know.

pyProbeParticle/ocl/AFMulator.py
ondrejkrejci: I see, good,

ondrejkrejci (approved):
After previous conversation approving now.

PR #37, opened 2022-10-28
Deleted old OCL related functions no longer in use.

### LINKED ISSUE #25: Delete all functions related to old procedural version of the OCL code.
Delete all functions related to old procedural version of the OCL code. If there is some useful functionality still there, it should be merged in to the object-oriented AFMulator API.

NikoOinonen:
Fixes #25

The deleted functions were either used nowhere or in one of the RunOpenCLbatch*.py scripts or GUImorse.py, which we were discussing should be deleted anyway. 

ocl.HighLevel.py did not have any functions in use except for genFFSampling and loadSpecies, which were already defined in common.py, so I deleted the whole module. I just had to modify common.loadSpecies a bit so that it checks if the file exists.


yakutovicha (approved):
Thanks, @NikoOinonen. LGTM!

PR #44, opened 2022-11-09
Readthedocs

### LINKED ISSUE #40: Get docs hosted on readthedocs.org
Now that we have a way of building the html pages with sphinx, we can get the docs hosted on readthedocs.org, so that they can be viewed easily

NikoOinonen:
Fixes #40

This adds configuration files for readthedocs so that the API docs will automatically get rebuilt on every commit to the main branch and the result gets hosted on a web page that is accessible to everyone. Also added a badge to the README that links to the latest version of the docs (which does not work yet, before this is merged). The current test version can be viewed at https://probeparticlemodel.readthedocs.io/en/readthedocs/.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 
 # Probe Particle Model (PPM)
 
+[![Documentation Status](https://readthedocs.org/projects/probeparticlemodel/badge/?version=main)](https://probeparticlemodel.readthedocs.io/en/main/?badge=main)
+
 Simple and efficient **simulation software for high-resolution atomic force microscopy** (**HR-AFM**) and other scanning probe microscopy (SPM) techniques with sub-molecular resolution (STM, IETS, TERS). It simulates deflection of the particle attached to the tip (typically CO molecule, but also e.g. Xe, Cl-, H2O and others).
 
 ### References
--- a/README.md
+++ b/README.md
@@ -1,13 +1,12 @@
 
 # Probe Particle Model (PPM)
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -1,6 +1,8 @@
 
 # Probe Particle Model (PPM)
 
+[![Documentation Status](https://readthedocs.org/projects/probeparticlemodel/badge/?version=main)](https://probeparticlemodel.readthedocs.io/en/main/?badge=main)
```
ondrejkrejci, line 4: Do we want to have this documentation for the whole code, which is absolutely not like 1.st meeting tutorial that high. I would put it on a same the same level as the inner wiki?!

ondrejkrejci (commented):
Except for the README.md I do not find any problems, But I find that README part important.

README.md
ondrejkrejci: Well, at least personally, I would slightly change the first sub-section of README called **References** to something like **Further information:** This one should have 3 points:
 - https://github.com/Probe-Particle/ProbeParticleModel#notable-publications-using-probe-particle-model Which points to literature at the end of README
 - https://probeparticlemodel.readthedocs.io/en/readthedocs/index.html 
 - https://github.com/Probe-Particle/ProbeParticleModel/wiki

Otherwise we have there the literature 2 times.

NikoOinonen: Sounds good. I changed it to look like this: https://github.com/Probe-Particle/ProbeParticleModel/blob/readthedocs/README.md. Would that be okay?

ondrejkrejci (approved):
Good.

yakutovicha:
Guys, please squash and merge 😄

PR #45, opened 2022-11-09
Master kpfm

### LINKED ISSUE #43: Merging the KPFM part to the main.
Aurelio will need to compare the latest development of KPFM in here https://github.com/aureliojgc/ProbeParticleModel/tree/OpenCL_KPFM_py3

aureliojgc:
Fixes #43 

Request for pull the kpfm modifications.
Basically I add the cycles over bias in generateELFF, both the relaxed_scan and relaxed_scan_PVE and the plot_results.
I made relaxed_scan_PVE to work also with LJ
The rest are minor changes around the kpfm needs.

I have tested all AFM and KPFM with density tips and orbital-like tips for AIMS and VASP.

### DIFF
--- a/pyProbeParticle/HighLevel.py
+++ b/pyProbeParticle/HighLevel.py
@@ -73,7 +73,7 @@ def relaxedScan3D( xTips, yTips, zTips, trj=None, bF3d=False ):
     if(verbose>0): print("<<<END: relaxedScan3D()")
     return fzs,PPpos
 
-def perform_relaxation (lvec,FFLJ,FFel=None, FFpauli=None, FFboltz=None,tipspline=None,bPPdisp=False,bFFtotDebug=False):
+def perform_relaxation (lvec,FFLJ,FFel=None, FFpauli=None, FFboltz=None,FFkpfm_t0sV=None,FFkpfm_tVs0=None,tipspline=None,bPPdisp=False,bFFtotDebug=False):
     if(verbose>0): print(">>>BEGIN: perform_relaxation()")
     if tipspline is not None :
         try:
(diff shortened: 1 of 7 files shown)


pyProbeParticle/HighLevel.py
```diff
@@ -212,7 +216,7 @@ def computeELFF_pointCharge( geomFile, tip='s', save_format=None, computeVpot=Fa
     if(verbose>0): print("<<<END: computeELFF_pointCharge()")
     return FF, V, nDim, lvec
 
-def computeElFF(V,lvec,nDim,tip,computeVpot=False, tilt=0.0 ):
+def computeElFF(V,lvec,nDim,tip,computeVpot=False, tilt=0.0,sigma=PPU.params['sigma'] ):
```
NikoOinonen, line 219: This is probably not doing what you think it does. The function signature gets evaluated only once when the module is imported, so the `sigma` value here will not get updated even if the value in `PPU.params['sigma']` changes. It's probably better if you set the default value to `None` and then inside function check if it's `None` and set it to `PPU.params['sigma']` if it is.

aureliojgc: Hi Nico,
Thanks for the review. I rewrote the function acording to your comments and tested it againg. I think it is correct now.

NikoOinonen: Ok, looks good.

ProkopHapala (approved):
seems fine, assuming you tested original functionality works (I didn't)

ProkopHapala (changes requested):
Seems fine, assuming you tested it. 

Consider if you really need to copy V and rho before you subtract it? Do you need to save original array? Are you worried that the original array may be changed? It can cost quite a lot of memory if the arrays are large.

ProkopHapala (approved):

ProkopHapala (approved):
seems fine assuming you tested that original functionality works (I didn't)

relaxed_scan.py
```diff
@@ -69,12 +73,26 @@ def rotFF( Fx,Fy, a ):
...
         if ( abs(Q) > 1e-7):
             charged_system=True
+    # Vkpfm
+    aplied_bias=False
+    if opt_dict['Vrange'] is not None:
+        Vs = np.linspace( opt_dict['Vrange'][0], opt_dict['Vrange'][1], int(opt_dict['Vrange'][2]) )
+    elif opt_dict['Vbias'] is not None:
+        Vs = [ opt_dict['Vbias'] ]
+    else:
+        Vs = [0.0]
```
ProkopHapala, line 83: OK, so it now always loop over Vs, but when KPFM is inactive Vs=[0.0]

ProkopHapala (approved):
seems OK, as long as it is tested

PR #49, opened 2022-11-23
Simplified xyz IO functions

### LINKED ISSUE #24: Single function for reading xyz and writing xyz files
In basUtils there are 4 different functions for reading xyz files that return results in slightly different formats and 6 different functions for writing xyz files. We should agree on single interface for these.

NikoOinonen:
Fixes #24 

Deletes all old xyz IO functions and replaces them with new ones called `loadXYZ` and `saveXYZ` in all relevant places. Also handle the extended xyz file format used by ASE, in order to pick out the correct column for charges and interpret the comment line for the lattice vectors.

### DIFF
--- a/pyProbeParticle/basUtils.py
+++ b/pyProbeParticle/basUtils.py
@@ -347,6 +347,9 @@ def loadGeometry(fname=None,params=None):
         elif not has_charges:
             # The fifth column in the ASE format was not charges, so ignore what was read from the file
             qs = np.zeros(len(qs), dtype=np.float32)
+        if _not_actually_charge(qs):
+            # qs is not actually charges based on some heuristics
+            qs = np.zeros(len(qs), dtype=np.float32)
         atoms = [list(Zs), list(xyzs[:, 0]), list(xyzs[:, 1]), list(xyzs[:, 2]), list(qs)]
     elif(is_cube):
         atoms = loadAtomsCUBE(fname)
@@ -392,6 +395,13 @@ def parseCommentASE(comment):
         has_charges = False
     
     return lvec, has_charges
+
+def _not_actually_charge(qs):
+    if abs(sum(qs)) > 3:
+        return True
+    if np.abs(qs).max() > 3:
+        return True
+    return False
 
 def findBonds( atoms, iZs, sc, ELEMENTS = elements.ELEMENTS, FFparams=None ):
     bonds = []
--- a/pyProbeParticle/basUtils.py
+++ b/pyProbeParticle/basUtils.py
@@ -4,8 +4,84 @@ from . import elements
 #import elements
 import math
 import numpy as np
+import warnings
 
 verbose = 0
+
+def loadXYZ(fname):
+    '''
+    Read the contents of an xyz file.
+
+    Arguments:
+        fname: str. Path to file.
+
+    Returns:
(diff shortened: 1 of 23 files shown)


pyProbeParticle/basUtils.py
```diff
@@ -1,131 +1,124 @@
...
+            qs = np.zeros(len(extra_cols), dtype=np.float64)
+    else:
+        # Not ASE format, so just take first column
+        qs = np.array([float(ex[0]) for ex in extra_cols], dtype=np.float64)
+        if _notActuallyCharge(qs):
+            # qs is not actually charges based on some heuristics
+            qs = np.zeros(len(qs), dtype=np.float64)
+    return qs
+
+def _notActuallyCharge(qs):
```
yakutovicha, line 91: I don't think the utility functions should check physics. This sort of goes against the separation of concern principle.

ondrejkrejci: Well, you never have systems, which would have abs(charges) higher than 3, 5 or 10. You would get completely crazy results and this could help you to debug the problem. I would higher the treshold for 10 here. Maybe @ProkopHapala you could also tell your opinion?

NikoOinonen: A better place for this check could be in `loadGeometry` after reading the xyz file?

yakutovicha: > Well, you never have systems, which would have abs(charges) higher than 3, 5 or 10. You would get completely crazy results and this could help you to debug the problem.

I am not questioning the need for such checks. They can be helpful. However, I would advocate for a separate module for such analysis.

My point is: why XYZ reader should be so complex? One day we want to replace the reader with something else, we need to monkey-patch all these checks. Why not already have a "check_physics" module that acts on ready-to-be-simulated structure? It could even be that I have an XYZ file with random charges that I am not even planning to use at all.

> A better place for this check could be in `loadGeometry` after reading the xyz file?

I would advocate for a separate module `check_physics`. It could even be kept inside utils - I don't have an opinion here.

NikoOinonen: > I would advocate for a separate module `check_physics`

That would make sense. But I think this would require more discussion about where should we use it and also if there are other checks that we want to make. I propose that we just delete the check from this PR and make a new issue to discuss it further, so that this PR can be merged.

ondrejkrejci: Ok, @yakutovicha - I got your point. @NikoOinonen - I agree - let's remove it now and put it as a separate function or class in the future.
I would then propose to have the check on point-charges somewhere in the _generateElFF_ ... part. So one do not follow the calculation with really wrong system.

NikoOinonen: Ok, I now removed the charge check, and made a new issue at #51.

yakutovicha: 👍

yakutovicha:
@NikoOinonen I think the branch should be rebased on top of `main`. Would you manage to do that? There are a lot of conflicts currently.

Once done, I will make a second pass.

NikoOinonen:
@yakutovicha Done rebasing.

yakutovicha (approved):
LGTM! Thanks a lot, @NikoOinonen!

PR #54, opened 2022-11-29
apos -> poss

alexriss:
seems to be some leftover from old code - with `apos` it doesn't compile, with `poss` it does.


NikoOinonen (approved):
I did not participate in developing this code, but this looks good to me.

It's a bit odd. It was changed from `poss` to `apos` in this [commit](https://github.com/Probe-Particle/ProbeParticleModel/commit/7ef9e6f412e54d7392b6754916678243b41b7cb1), but as far as I can tell the [actual attribute](https://github.com/Probe-Particle/ProbeParticleModel/blob/34e6a3603bfe4759fa68a0641ae202c30f64e778/cpp/RARFFarr.h#L184) was never called `apos`.

PR #56, opened 2022-11-30
Add initial Docker setup.

yakutovicha:


ondrejkrejci (approved):
The _python3.6_ in Dockerfile is the lowest possible version?

ondrejkrejci:
Sorry, I was too fast today :-).

pyproject.toml
```diff
@@ -20,8 +20,8 @@ requires-python = ">=3.6"
 dependencies = [
     "ase",
     "matplotlib",
-    "numpy",
-    "pyopencl",
+    "numpy<=1.21.6",
+#    "pyopencl",
```
NikoOinonen, line 24: Why the limit on numpy version and removed dependency on pyopencl?

Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM ubuntu:22.10
+
+RUN apt-get update && apt-get install -y \
+    build-essential                      \
+    python3.6                            \
```
NikoOinonen, line 5: I'm confused about this package. I'm pretty sure this does not exist in Ubuntu 22.10, only 18.04: https://packages.ubuntu.com/search?keywords=python3.6. Also, when I tried building the docker image, the version of python that I got was 3.10.

README.md
```diff
@@ -112,5 +112,22 @@ Hover the mouse cursor over any parameter for a tooltip explaining the meaning o
 * [Prokop Hapala, Georgy Kichin, Christian Wagner, F. Stefan Tautz, Ruslan Temirov, and Pavel Jelínek, Mechanism of high-resolution STM/AFM imaging with functionalized tips, Phys. Rev. B 90, 085421 – Published 19 August 2014](http://journals.aps.org/prb/abstract/10.1103/PhysRevB.90.085421)
 * [Prokop Hapala, Ruslan Temirov, F. Stefan Tautz, and Pavel Jelínek, Origin of High-Resolution IETS-STM Images of Organic Molecules with Functionalized Tips, Phys. Rev. Lett. 113, 226101 – Published 25 November 2014,](http://journals.aps.org/prl/abstract/10.1103/PhysRevLett.113.226101)
 
+## Making ppafm platform-independent.
+
+We propose to use [Docker](https://docs.docker.com/get-docker/) to make the code platform-independent.
```
ondrejkrejci, line 117: I am puzzled now. I have originally thought, that it would work within **pip** or **conda** environment and work without any compilation (for non-developers), like _ase_ is working. I think, the best think is to speak today.

ondrejkrejci (commented):
Let's speak today ...

Dockerfile
yakutovicha: Agreed, replaced with `python3` in 647d63600d667e1d561a217235504667df1d513f

pyproject.toml
yakutovicha: Added a comment on numpy upper limit, moved pyopencl to optional dependencies.

README.md
yakutovicha: We've discussed this today during the meeting, but just for the record:

- The app should still be installable through pip, and this remains our high priority.
- Docker is just an alternative way to deploy `ppafm`, which does not exclude the other options.

yakutovicha:
In my opinion, the PR is ready. Please have a look again and let me know if you approve it.

NikoOinonen (approved):
Ok, looks good to me.

Issue #31, opened 2022-10-25
Agree on the released name.


yakutovicha:

NikoOinonen:
I think @ondrejkrejci was suggesting in the meeting `ppafm`, which I like. It's short and to the point.

yakutovicha:
> I think @ondrejkrejci was suggesting in the meeting `ppafm`, which I like. It's short and to the point.

I agree. Anyone else has different ideas? @ProkopHapala, @mondracek?

yakutovicha:
Ok, if there are no other opinions, we stick with @ondrejkrejci's suggestion to call it `ppafm`.

PR #57, opened 2022-11-30
Provide `pypackage.toml`, rename root module to `ppafm`

### LINKED ISSUE #32: Organize the repository according to the python packaging guidelines.

yakutovicha:
fixes #32 

Here is the discussion for the package name agreement: #31


yakutovicha:
Just in case: despite the large number of files affected, the number of changes is not that big:

- I renamed the root package folder to `ppafm`
- I added `pyproject.toml` to make ppafm a package.
- I changed all `pyProbeParticle` imports to `ppafm` imports (this was done automatically).

ondrejkrejci (approved):
Please wait @yakutovicha - unless we are really in hurry;
Please @NikoOinonen - I have renamed all the _rst_ files for the documentation. Please have a look at those, so it would work. Thank you!

yakutovicha:
> Please wait @yakutovicha - unless we are really in hurry;

I have time :)

pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+license = {file = "LICENSE"}
+classifiers = [
+    "Development Status :: 3 - Alpha",
+    "Intended Audience :: Science/Research",
+    "Natural Language :: English",
+    "License :: OSI Approved :: MIT License",
+    "Programming Language :: Python :: 3",
+]
+requires-python = ">=3.6"
+dependencies = [
```
NikoOinonen, line 20: We also depend on `matplotlib` in the main code, in the OpenCL code we depend on `pyopencl` and `reikna`, and for the GUI there is an optional dependency on `ase` for viewing the molecule geometry.

pyproject.toml
```diff
@@ -0,0 +1,43 @@
+[build-system]
+requires = ["setuptools>=61.2"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "ppafm"
+dynamic = ["version"]
+description = "Classical forcefiled model for simulation atomic force microscopy images."
```
NikoOinonen, line 8: Let's fix the bad English here. I would change that to "Classical force field model for simulating atomic force microscopy images."

I would also change the description on the main page, but I don't think I have the rights to do that.

NikoOinonen (changes requested):
A couple of small things to address. @ondrejkrejci Thanks, the documentation works.

pyproject.toml
yakutovicha: done

pyproject.toml
yakutovicha: Ok, I will put everything into the `dependencies`. We can split those things later if needed.

NikoOinonen (approved):
Ok, looks good.

Issue #42, opened 2022-11-07
GUI enhancements


NikoOinonen:
The GUI is pretty well functioning now, but I have noted some features missing or non-optimal, so I thought to just list them here so that they can be worked on at some point.

- [ ] Tip density for point-charge systems
  - Currently, if you give point charge inputs, the tip is point charge as well. There should be an option to use the multipole tip densities for point-charge inputs as well.
- [ ] Rotations for Hartree potential inputs and point-charge tip
  - Currently, if you give Hartree potential input (.xsf etc.) and try to use point-charge tip with sample rotation, it gives an error, because it's not implemented.
- [ ] Don't immediately update when typing values
  - The plot updates immediately after any change to the input boxes, which is annoying when you have not finished typing the value you want. It should function in a way that it only updates after pressing Enter after you typed a new value. However, it should still update immediately when using arrow keys or scroll wheel to change the values.
- [ ] Add a button to the force-field viewer window for saving the force field to a file.
- [ ] Memory management
  - The program does not keep any track of the used device memory, and it's very easy to make it crash by just increasing the scan size enough. There should be system for tracking the currently used memory compared to the total available and block any actions that would increase it over the maximum.
  - Show the user how much memory is in use, and display some kind of warning when it's close to the upper limit.
  - Some limit on the initial scan size so that it can't overflow the available memory.
- [ ] Some way of saving/loading state
  - When you are analysing some system, there should be some way of saving the current parameters to a file and then load them later. Maybe the parameters could be saved/loaded in the params.ini format, so that it would be interoperable with the CLI scripts.

- [ ] Optimizations:
  - Matplotlib is slow for real-time rendering. Updating the plot takes ~50-100ms, which is a major bottleneck especially for point-charge systems where the simulation only takes ~10-20ms. Should look into switching to using some other library.
  - Executing the simulations and updating the plot should be asynchronous. The GUI is unresponsive when the simulation is in progress, which becomes a problem with bigger systems.
  - Changing force-field box size with Hartree potential inputs is slow for two reasons:
    - The multipole tip density is currently calculated on the host and then uploaded to the device. This should all be done on the device.
    - The FFT for the tip convolution is implemented using the reikna library, which does a JIT compilation of the operation, which becomes quite slow for bigger grid sizes.

ondrejkrejci:
One comment
[...]

PR #67, opened 2022-12-15
GUI fixes

NikoOinonen:
A couple of small fixes for issues mentioned in #42 
- Default save path for output images and data is in the same directory as the input file.
- Force field viewer can save the force field to a file.
- Added a .desktop file for GUI for better integration with the desktop environment.
- Typing values does not update the plot before pressing return.
- General code clean-up.


README.md
```diff
@@ -78,11 +78,17 @@ Additionally an OpenCL Installable Client Driver (ICD) for your compute device i
 
 Run the GUI application:
 ```sh
-./ppm-gui
+./GUI/ppm-gui
+```
+
+In order to make the GUI application appear in the system application menu and 'Open with' context menus, link the `ppm-gui` application to a location that is on PATH, e.g. `~/.local/bin`, and install the provided .desktop file. This can be achived by running the following in the repository root:
+```bash
+ln -s `realpath ./GUI/ppm-gui` $HOME/.local/bin
```
ondrejkrejci, line 86: I guess, that will change, once we will have the full pip way, but OK for now.

ppafm/GUIWidgets.py
```diff
@@ -413,22 +419,27 @@ def __init__(self, parent=None, title="View Forcefield", width=5, height=4, dpi=
...
         sl.setCurrentIndex(sl.findText(components[2])); sl.currentIndexChanged.connect(self.updateView)
         sl.setToolTip(tooltips[0]); hbl.addWidget(sl)
-        hbl.addStretch(1)
 
+        hbr = QtWidgets.QHBoxLayout(); hb.addLayout(hbr)
         lb = QtWidgets.QLabel('z index:'); lb.setToolTip(tooltips[1]); hbr.addWidget(lb)
         bx = QtWidgets.QSpinBox(); bx.setRange(0, 200); bx.setValue(0)
         bx.valueChanged.connect(self.updateView); bx.setToolTip(tooltips[1])
         hbr.addWidget(bx); self.bxInd = bx
+        hbr.setContentsMargins(10, 0, 10, 0)
```
ondrejkrejci, line 434: Why such a chages in Margins?

ppafm/GUIWidgets.py
```diff
@@ -459,3 +470,35 @@ def updateView(self):
 
         if self.verbose > 0: print('FFViewer.updateView', ic, iz, data.shape)
         if self.verbose > 1: print('updateView time [s]', time.perf_counter() - t0)
+
+    def saveFF(self):
```
ondrejkrejci, line 474: This is great, and very usefull!

ppafm/GUIWidgets.py
```diff
@@ -413,22 +419,27 @@ def __init__(self, parent=None, title="View Forcefield", width=5, height=4, dpi=
...
         hb = QtWidgets.QHBoxLayout(); self.centralLayout.addLayout(hb)
-        hbl = QtWidgets.QHBoxLayout(); hb.addLayout(hbl)
-        hbr = QtWidgets.QHBoxLayout(); hb.addLayout(hbr)
 
+        hbl = QtWidgets.QHBoxLayout(); hb.addLayout(hbl)
         lb = QtWidgets.QLabel('Component:'); lb.setToolTip(tooltips[0]); hbl.addWidget(lb)
         sl = QtWidgets.QComboBox(); self.slComponent = sl; sl.addItems(components)
         sl.setCurrentIndex(sl.findText(components[2])); sl.currentIndexChanged.connect(self.updateView)
         sl.setToolTip(tooltips[0]); hbl.addWidget(sl)
-        hbl.addStretch(1)
```
ondrejkrejci, line 423: I hope that you know, what you are doing here.

GUI/ppm-gui
```diff
@@ -823,13 +641,293 @@ class ApplicationWindow(QtWidgets.QMainWindow):
...
+
+        # Left-right divide for labels and input boxes
+        vb = QtWidgets.QHBoxLayout(); self.l0.addLayout(vb)
+        bxl = QtWidgets.QVBoxLayout(); vb.addLayout(bxl, 1)
+        bxr = QtWidgets.QVBoxLayout(); vb.addLayout(bxr, 3)
+
+        # Lattice vector A
+        lb = QtWidgets.QLabel("A (x, y, z) [Å]"); lb.setToolTip(TTips['PBC']); bxl.addWidget(lb)
+        vb = QtWidgets.QHBoxLayout(); bxr.addLayout(vb)
+        self.bxPBCAx = _spin_box((-100, 100), 50, 0.1, self.updatePBC, TTips['PBC'], vb)
```
ondrejkrejci, line 787: Let's hope that no one will have bigger systems than 100angstrom in near future!

ondrejkrejci (approved):
Just few super-small comments inside, otherwise I do not find any other problems.

README.md
NikoOinonen: Yes, this is meant to be temporary.

ppafm/GUIWidgets.py
NikoOinonen: This and the following margin change were just to make it look a bit better with the addition of the button for saving the force field.

ppafm/GUIWidgets.py
NikoOinonen: See the other comment.

GUI/ppm-gui
NikoOinonen: Well, it's probably pretty rare to go above 100 Å, but I could just increase it up to 1000 Å.

ondrejkrejci: I do not think so for now, on most of nowadays machines, such a big FF would lead to _memory overflow_ ...

NikoOinonen: There could be cases where the periodic cell is big in one direction, but narrow in another, like some ribbons of 2D materials. Also, the lattice vector size here does not affect the memory usage. I think pyqt just requires the maximum values to determine the minimum size of the spin box. So there should be no harm in increasing the maximum value.

PR #87, opened 2023-02-14
Prokop damp func

### LINKED ISSUE #48: Choose damping function for vdW
It was noted in #35 that the stand-alone vdW calculation has some unusual term, which was used as a quick solution for damping the force to prevent instability. There exist better functions for the damping, e.g. https://manual.q-chem.com/5.2/Ch5.S7.SS2.html. We should choose one that has a good compromise between stability and speed.

ProkopHapala:
fixes #48 

So I implemented 4 other damping function. In the process of testing I also corrected bug in PyridineDensOverlap example.

@NikoOinonen : When I plot the new damped vdW potential in 3D as isosurface in VESTA it still does not look like the atoms, just as a blob, but I think it is fine. Actually I think the old daping (with constant) was not that bad (considering how simple it was). 

In general the main question is how to choose the parameter ADamp (streght of damping). It ADamp is large the damping affect the potential at r=Rmin, if ADamp is small the potential is deep and it can cause the probe particle to fall inside atom, which is unrealistic. But in order to optimize ADamp properly it would requite e.g. to fit it with respect to DFT calculations (which is actually what Ruben Perez did).

### DIFF
--- a/cpp/Forces.h
+++ b/cpp/Forces.h
@@ -211,7 +211,7 @@ double addAtomVdW_addDamp( const Vec3d& dR, Vec3d& fout, double R, double E0, do
     return E;
 }
 
-double R2_func(double r2,double &df){
+inline double R2_func(double r2,double &df){
     //printf( "R2_func() \n");
     if(r2>1){
         df     =  0;
@@ -222,7 +222,7 @@ double R2_func(double r2,double &df){
     }
 }
 
-double R4_func(double r2,double &df){
+inline double R4_func(double r2,double &df){
     //printf( "R4_func() \n");
     if(r2>1){
         df     =  0;
@@ -234,7 +234,7 @@ double R4_func(double r2,double &df){
     }
 }
 
-double invR4_func(double r2,double &df){
+inline double invR4_func(double r2,double &df){
     //printf( "invR4_func() \n");
     double invR2 = 1/r2;
     double invR4 = invR2*invR2;
@@ -242,7 +242,7 @@ double invR4_func(double r2,double &df){
     return            invR4;
 }
 
-double invR8_func(double r2,double &df){
+inline double invR8_func(double r2,double &df){
     //printf( "invR8_func() \n");
     double invR2 = 1/r2;
     double invR8 = invR2*invR2; invR8*=invR8;
--- a/cpp/Forces.h
+++ b/cpp/Forces.h
@@ -122,12 +122,81 @@ inline double addAtomLJ( const Vec3d& dR, Vec3d& fout, double c6, double c12 ){
 }
 
 // Lenard-Jones force between two atoms a,b separated by vector dR = Ra - Rb
-inline double addAtomVdW( const Vec3d& dR, Vec3d& fout, double c6 ){
+inline double addAtomVdW_dampConst( const Vec3d& dR, Vec3d& fout, double c6 ){
     double r2 = dR.norm2(); r2*=r2; r2*=r2;
     //fout.add_mul( dR , 6*c6 /( r2 + 1.0 ) );
     //fout.add_mul( dR , 6*c6 /( r2 + 60*c6 ) );
     fout.add_mul( dR , 6*c6 /( r2 + 180*c6 ) );
     return 0;
+}
+
+
+
+/*
+// Lenard-Jones force between two atoms a,b separated by vector dR = Ra - Rb
+inline double addAtomVdW_dampR4( const Vec3d& dR, Vec3d& fout, double E0, double invR2cut, double cdamp ){
+    double r2 = dR.norm2(); 
+    double r8=r2*r2; r8=r8*r8;
+    double R2fun = 1-r2*invR2cut;
+    double dR2   = 1-r2*invR2cut;
+    if (damp<0) damp=0;  
+    F = Eij*r*( -C6/(r8 + cdamp*R2fun*R2fun  )  )  
+    return 0;
+}
+
+// https://manual.q-chem.com/5.2/Ch5.S7.SS2.html
+inline double addAtomVdW_Chai( const Vec3d& dR, Vec3d& fout, double E0,  double invR2cut, double cdamp ){
+    double r2 = dR.norm2()*invR2cut; 
+    double r4=r2*r2; r8=r8*r8;
+    double d  = 1/r8;
+    double D = 1./( 1. + A * D );
+    double D = 8*A*D*E*E/r**2
+
+    F = 12*A*D*E*E/r**2
+
+    double damp = 1-r2*invR2cut;  // damp = 1-(r/Rcut)^2 
+    if (damp<0) damp=0;  
+    F = Eij*r*( -C6/(r8 + cdamp/r8 )  )  
+    return 0;
+}
+*/
+
+template<double Rfunc(double r2,double &df)>
+double addAtomVdW_addDamp( const Vec3d& dR, Vec3d& fout, double R, double E0, double ADamp ){
+  double D,dD;
+  double r2 = dR.norm2()/(R*R); 
+  double r4 = r2*r2; 
+  D  = Rfunc(r2,dD);
+  double e  = 1./(      r4*r2 + D*ADamp);                    
+  double E  = -E0*e;
+  double fr = - E*e*( 6*r4    + dD*ADamp ); 
+  fout.add_mul(dR,fr);
+  return E;
+}
+
+double R2_func(double r2,double &df){
+    df     =  -2;
+    return    1 - r2;
+}
+
+double R4_func(double r2,double &df){
+    double e = 1 - r2; 
+    df     =  -4*e;
+    return     e*e;
+}
+
+double invR4_func(double r2,double &df){
+    double invR2 = 1/r2;
+    double invR4 = invR2*invR2;
+    df           = -4*invR4*invR2;
+    return            invR4;
+}
+
+double invR8_func(double r2,double &df){
+    double invR2 = 1/r2;
+    double invR8 = invR2*invR2; invR8*=invR8;
+    df           = -8*invR8*invR2;
+    return            invR8;
 }
 
 // Morse force between two atoms a,b separated by vector dR = Ra - Rb
--- a/cpp/Forces.h
+++ b/cpp/Forces.h
@@ -149,50 +149,6 @@ inline double addAtomVdW_dampConst( const Vec3d& dR, Vec3d& fout, double c6, dou
     return 0;
 }
 
-
-
-/*
-// Lenard-Jones force between two atoms a,b separated by vector dR = Ra - Rb
-inline double addAtomVdW_dampR4( const Vec3d& dR, Vec3d& fout, double E0, double invR2cut, double cdamp ){
-    double r2 = dR.norm2(); 
-    double r8=r2*r2; r8=r8*r8;
-    double R2fun = 1-r2*invR2cut;
-    double dR2   = 1-r2*invR2cut;
-    if (damp<0) damp=0;  
-    F = Eij*r*( -C6/(r8 + cdamp*R2fun*R2fun  )  )  
-    return 0;
-}
-
-// https://manual.q-chem.com/5.2/Ch5.S7.SS2.html
-inline double addAtomVdW_Chai( const Vec3d& dR, Vec3d& fout, double E0,  double invR2cut, double cdamp ){
-    double r2 = dR.norm2()*invR2cut; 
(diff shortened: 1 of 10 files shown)


cpp/Forces.h
```diff
@@ -122,13 +134,121 @@ inline double addAtomLJ( const Vec3d& dR, Vec3d& fout, double c6, double c12 ){
...
+    double D,dD;
+    double r2    = dR.norm2();
+    double invR2 = 1./(R*R);
+    double u2    = r2*invR2; 
+    double u4    = u2*u2; 
+    D  = Rfunc(u2,dD);
+    //ADamp=0; // DEBUG
+    double e  = 1./(      u4*u2 + D*ADamp);       
+    //printf( "r %g e %g R0 %g E0 %g \n", sqrt(r2), e,  R, E0 );             
+    double E  = -2*E0*e;
```
NikoOinonen, line 207: I have usually seen the energy parameter with prefactor 4 instead of 2, like on wikipedia: https://en.wikipedia.org/wiki/Lennard-Jones_potential. Or maybe the parameter is defined somehow differently here?

ProkopHapala: It depends if you use `(sigma/r)^6` (sigma is position of node) or `(r0/r)^6` (r0 is position of minimum) 

see 
n-exp form in
https://en.wikipedia.org/wiki/Lennard-Jones_potential#Alternative_notations_of_the_Lennard-Jones_potential

cpp/Forces.h
```diff
@@ -122,13 +134,121 @@ inline double addAtomLJ( const Vec3d& dR, Vec3d& fout, double c6, double c12 ){
...
+    double fr = -E *e*( 6*u4    + dD*ADamp)*invR2 ; 
+    //fr = dD*E0; E = D*E0; // DEBUG
+    fout.add_mul(dR,fr);
```
NikoOinonen, line 210: I'm not convinced that this formula for the force is correct. From my calculation, the exponent for `u` in the numerator should be 5 instead of 4, and to make the unit vector for `dR` I think the denominator should be `sqrt(r2)`, so something like
```C
double fr = -E *e*( 6*u4*sqrt(u2) + dD*ADamp) / sqrt(r2);
```
Also, is the sign correct? The force is the negative derivative of the energy, but you also get one minus sign from the derivative of the denominator.

cpp/Forces.h
```diff
@@ -122,13 +134,121 @@ inline double addAtomLJ( const Vec3d& dR, Vec3d& fout, double c6, double c12 ){
...
-    //fout.add_mul( dR , 6*c6 /( r2 + 1.0 ) );
-    //fout.add_mul( dR , 6*c6 /( r2 + 60*c6 ) );
-    fout.add_mul( dR , 6*c6 /( r2 + 180*c6 ) );
+    fout.add_mul  ( dR , 6*c6 /( r2 + ADamp*c6 ) );
+    return 0;
+}
+
+
+
+/*
```
NikoOinonen, line 154: Do these commented-out blocks serve a purpose? I would just delete them to make it cleaner, if they are not required for later.

generateElFF.py
```diff
@@ -59,7 +59,14 @@
     else:
         FFparams = PPU.loadSpecies( cpp_utils.PACKAGE_PATH+'/defaults/atomtypes.ini' )
 
-    if ( (options.doDensity) and (options.Rcore > 0.0) and (options.tip is None) ):  # We do it here, in case it crash we don't want to wait for all the huge density files to load
+
+    print( ">>>>DEBUG !!!!!!!!!" )
+    print(  "\ndoDensity ",options.doDensity, "\nRcore ", options.Rcore,  "\ntip ", options.tip,  "\ntip_dens ", options.tip_dens )
+    print( "<<<<DEBUG !!!!!!!!!" )
```
NikoOinonen, line 65: Remove the debug prints before merge.

cpp/ProbeParticle.cpp
```diff
@@ -357,6 +367,45 @@ void getCoulombFF( int natoms_, double * Ratoms_, double * kQQs, int kind ){
...
+            case  3: if(ADamp_>0){ ADamp_invR4 = ADamp_; }; E=addAtom_VdW_invR4( dR, fout, coefs ); break;
+            case  4: if(ADamp_>0){ ADamp_invR8 = ADamp_; }; E=addAtom_VdW_invR8( dR, fout, coefs ); break;
+        }
+        Fs[i]=fout.x;
+        Es[i]=E;
+    }
+}
+
+void getVdWFF_RE( int natoms_, double * Ratoms_, double * REs, int kind, double ADamp_=-1.0 ){
+    printf( "DEBUG getVdWFF_RE(kind=%i,ADamp=%g) \n", kind, ADamp_ );
```
NikoOinonen, line 392: Remove the debug print before merge.

cpp/Forces.h
NikoOinonen: Ok, just wanted to make sure it's consistent.

ProkopHapala: yes, I think it is good you check it. Because I often do mistakes of this king. So does it seems correct to you now?

cpp/ProbeParticle.cpp
ProkopHapala: OK, I removed the debug prints

cpp/Forces.h
NikoOinonen: I think that would be true without the damping function, but you are also multiplying the damping function by `r` which would cause some distortion to the damping of the force.

`invR2` here is the inverse of the square of the LJ parameter, isn't it, so it's not the same as `1 / sqrt(r2) = 1 / |dR|` which I thought of as the normalization for `dR` as you said.

For the minus signs, I think there should be three, one for the energy, one for the fact that `F = -dE/dr`, and one from the derivative of `1/f(r)`. But I think you only have two, because one is already in the energy, and the other two should cancel out in the force formula.

ProkopHapala: * Notice that  dD returned from Rfunc is also missing one power of `r` exactly for this reason that it is expected to be multiplied by `dR`. Therefore dD = (d Rfunc(r)/dr)/r` rater than just `d Rfunc(r)/dr`
* you are right invR2 si 1/R0^2 not 1/r^2, I overlooked that while replying you. But still the code is fine (I think):

```
E(r) = 1/( r^6 + Rfunc(r) )
F(r) =  dR *   -E(r)^2 * ( 6*r^4 + (dRfunc(r)/dr)/r )
```

* ad sign - I always make mistakes in this. Never shure if we store `F=-dE/dr` or `F=dE/dr`. But notice that the test here works (meaning the sign convecction is the same as for un-damped L-J and vdW )
https://github.com/Probe-Particle/ppafm/issues/48#issuecomment-1429919012

please try tu run the test here if you don't believe me:
https://github.com/Probe-Particle/ppafm/blob/prokop-damp-func/dev/test_vdwDaping.py

NikoOinonen: Ah, I didn't see that `dD` is missing a factor of `r`. Then it makes sense. I'm still a bit confused by the sign convention, but it looks like it comes out correct in the end, so I guess it's fine.

NikoOinonen (approved):
Looks good to me now.

ProkopHapala:
OK, @NikoOinonen reviewed it quite troughfully. Therefore, should I wait for @ondrejkrejci  @yakutovicha or should I merge it straightaway ?

yakutovicha (approved):
No issues from my side, feel free to merge that

generateElFF.py
```diff
@@ -95,7 +97,7 @@
         print(">>> loading tip density from ",options.tip_dens,"...")
         rho_tip, lvec_tip, nDim_tip, head_tip = GU.loadXSF( options.tip_dens )
```
ondrejkrejci, line 98: What if tip is calcualated by aims or cp2k? Some more general loading would be better, at least in the future.

ppafm/defaults/params.ini
```diff
@@ -33,3 +33,4 @@ gridC            0        0      5.0    # c-vector of unit cell; recomanded form
 scanMin      0.0    0.0    5.0          # start of scanning (x,y,z)
 scanMax     20.0   20.0    8.0          # end of scanning (x,y,z)
 Amplitude       1.0						# [Å] oscilation amplitude (peak-to-peak) for conversion Fz->df
+vdWDampKind   2                         # vdW damping 0:Constant, 1:R2func, 2:R4func, 3:invR4, 4:invR8
```
ondrejkrejci, line 36: Once more here it is writing me no free line, but I have a feeling that a free line would actually make problems.

ondrejkrejci (approved):
All the comments are minor, thus I approve. Still my comments to [examples/pyridineDensOverlap/run.sh-](https://github.com/Probe-Particle/ppafm/pull/87/files#diff-8f87b7d6c52495365919e44e7ee53aa3afcea7eae88c0fd45d7d7c29bc0b94c5) are worth of checking.

cpp/ProbeParticle.cpp
```diff
@@ -357,6 +367,45 @@ void getCoulombFF( int natoms_, double * Ratoms_, double * kQQs, int kind ){
...
+            case  4: if(ADamp_>0){ ADamp_invR8 = ADamp_; }; E=addAtom_VdW_invR8( dR, fout, coefs ); break;
+        }
+        Fs[i]=fout.x;
+        Es[i]=E;
+    }
+}
+
+void getVdWFF_RE( int natoms_, double * Ratoms_, double * REs, int kind, double ADamp_=-1.0 ){
+    //printf( "DEBUG getVdWFF_RE(kind=%i,ADamp=%g) \n", kind, ADamp_ );
+    natoms=natoms_; Ratoms=(Vec3d*)Ratoms_; nCoefPerAtom = 2;
```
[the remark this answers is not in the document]
ProkopHapala: it is not used in this function, it is global variable used for walking over coef array in [evalCell()](https://github.com/Probe-Particle/ppafm/blob/4598d7d56f7cf04d3004714ed950e68d689c7889/cpp/ProbeParticle.cpp#L220)

it is not very readable and not-good practice, but it allows to use generic function to evaluate all interactions

dev/test_vdwDaping.py
```diff
@@ -0,0 +1,105 @@
...
+    Fs_num,xs_  = deriv( xs, Es )
+
+    plt.subplot(2,nz,i   +1);    plt.plot(xs,Es_LJ   ,':k'); plt.plot(xs,Es_r6   ,'--k'); plt.plot(xs,Es   );                                   plotDecor( R0, E0*5 );  plt.title("%s(Adamp=%5.2f) %s-%s" %(name,ADamp, (FFparams[iZs[i]-1][4]).decode(), (FFparams[iPP-1][4]).decode() ) ) #; plt.title("Interaction %i-%i" %(iZs[i], iPP) )
+    plt.subplot(2,nz,i+nz+1);    plt.plot(xs,Fs_LJ*-1,':k'); plt.plot(xs,Fs_r6*-1,'--k'); plt.plot(xs,Fs*-1);  plt.plot(xs_,Fs_num*-1, ':r');   plotDecor( R0, E0*7 ); 
+
+plt.subplot(2,nz,  +1); plt.ylabel("Energy [eV]");
+plt.subplot(2,nz,nz+1); plt.ylabel("Force  [eV/A]");
+plt.savefig(name+".png",bbox_inches='tight')
+
+plt.show()
```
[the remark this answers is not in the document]
ProkopHapala: I would keep that for formater

examples/pyridineDensOverlap/run.sh-
```diff
@@ -11,29 +11,28 @@ echo "in order to download data files we need to install megatools "
 echo "alternatively you may download it manually from mega.nz webpage and place it to proper sub-folders ./tip and ./sample "
 echo "If you have Debian based linux system with apt-get this should work : "
 echo "enter your sudo password ( I promiss not hacking you ;-) : "
-#sudo apt-get install megatools
+sudo apt-get install megatools
 
-echo "Downloading Sample data-files ... "
-#megadl 'https://mega.nz/#F!GOhnRIwY!XneVlLYhCmvp74JCCH9BRA'
+#echo "Downloading Sample data-files ... "
```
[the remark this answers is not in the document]
ProkopHapala: this is link to the original file which is missing so I replaced by new links

generateElFF.py
ProkopHapala: Yes, but keep this for future (I do not have time to test it now)

yakutovicha:
@ProkopHapala, please squash-merge

Issue #76, opened 2023-02-09
Using the aliases of builtin types like `numpy.int` is deprecated


yakutovicha:
In #56 I had to fix the upper version of numpy (see the [numpy deprecation note](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations)) due to the removal of `numpy.int`, `numpy.float`, etc.

We need to fix the code according to the [guidelines](https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated) and unblock the version again.

NikoOinonen:
I can take care of this.

PR #91, opened 2023-02-20
starting to work on #53;

### LINKED ISSUE #53: Saving the LJ forcefield in numpy format with generateLJFF.py raises error
While trying to do relaxed scans with PPSTM, I found that generating the LJ forcefield with `python generateLJFF.py -i INPUT_FILE.xyz -f npy` seems to be broken at the moment in the main branch, as there is an `IndexError: string index out of range` raised by line 390 in the attached snippet . 

https://github.com/Probe-Particle/ProbeParticleModel/blob/34e6a3603bfe4759fa68a0641ae202c30f64e778/pyProbeParticle/GridUtils.py#L382-L390

This happens because `head` is a string here, looks like something is wrong either in the indexing or the definition of `head`? Ultimately the variable `head` is defined here as `atomstring`:

https://github.com/Probe-Particle/ProbeParticleModel/blob/34e6a3603bfe4759fa68a0641ae202c30f64e778/pyProbeParticle/HighLevel.py#L150

So the question is: does this function work properly or is there something wrong in the atom-writing function in GridUtils.py?

PS. This feature of saving atoms was not implemented in the old main branch of ProbeParticleModel and generating the force field using the `master_backup` branch works fine. 

Best,
Lauri

ondrejkrejci:
fixes #53
adding there atoms [e,x,y,z] array into GridUtils and HighLevel
**In Progress**

### DIFF
--- a/ppafm/GridUtils.py
+++ b/ppafm/GridUtils.py
@@ -337,14 +337,16 @@ def saveWSxM_3D( prefix, data, extent, slices=None ):
 
 #================ Npy
 
-def saveNpy(fname, data, lvec , head=None):
+def saveNpy(fname, data, lvec , atoms, head=None):
 	np.save(fname+'.npy', data)
 	np.save(fname+'_vec.npy',lvec)
+	np.save(fname+'_atoms.npy',np.array(atoms,dtype=float))
 
 def loadNpy(fname):
 	data = np.load(fname+'.npy')
 	lvec = np.load(fname+'_vec.npy')
-	return data.copy(), lvec;	#necessary for being 'C_CONTINUOS'
+	atoms = np.load(fname+'_atoms.npy') 
+	return data.copy(), lvec, atoms;	#necessary for being 'C_CONTINUOS'
 
 # =============== Vector Field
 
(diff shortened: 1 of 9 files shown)


ondrejkrejci:
Hei everyone; As I am getting through scripts like _generaleElFF.py_ I have noticed, that rhoTip is always automatically created and saved. I do not think, that this is necessary, that is why, I will add there an option to save it if needed, but otherwise, it will not save it ....

ondrejkrejci:
passing the atoms information through the script will follow ... hopefully tomorrow

ondrejkrejci:
While pushing to my branch,, I've got this comment:
```
remote: Resolving deltas: 100% (6/6), completed with 6 local objects.
remote: 
remote: GitHub found 1 vulnerability on Probe-Particle/ppafm's default branch (1 moderate). To find out more, visit:
remote:      https://github.com/Probe-Particle/ppafm/security/dependabot/1
remote: 
```
Maybe something we should also discuss ...

NikoOinonen:
> GitHub found 1 vulnerability on Probe-Particle/ppafm's default branch (1 moderate)

This is related to #76. We to fix some things to support the latest versions of numpy so that we can get rid of the dependency on the old version.

ondrejkrejci:
@NikoOinonen @yakutovicha and @ProkopHapala , you can have a look now. I have chosen the lowest drag way. which is maybe not that neat, but it is working and it is propagating the important information through. at the moment, I have things still in separate files in _NPY_ way, but we can discuss, whether we can pack it into the new **npz** format.

ondrejkrejci:
Maybe wait, I have forgotten to add the legend to some of the GridUtils functions, that I have worked on. I will hopefully work on it in the afternoon.

ppafm/basUtils.py
```diff
@@ -404,6 +404,31 @@ def parseLvecASE(comment):
         lvec = None
     return lvec
 
+def create_at_array(atoms,lvec):
```
yakutovicha, line 407: I don't think this function is needed. Its functionality can be replaced by one line:
```python
atoms_array = np.append(np.array(atoms), np.array(lvec))
```

ppafm/GridUtils.py
```diff
@@ -371,24 +391,22 @@ def loadVecFieldNpy( fname, FF = None ):
...
+def saveVecFieldNpy( fname, FF, lvec , at_array ):
 	np.save(fname+'_x.npy', FF[:,:,:,0] )
 	np.save(fname+'_y.npy', FF[:,:,:,1] )
 	np.save(fname+'_z.npy', FF[:,:,:,2] )
 	np.save(fname+'_vec.npy', lvec )
-	if (head != XSF_HEAD_DEFAULT ):
-		print("saving atoms")
-		tmp0=head[0]; q=np.zeros(len(tmp0));    #head: [e,[x,y,z],lvec]
-		np.save(fname+'_atoms.npy',[tmp0,head[1][0],head[1][1],head[1][2],q]) #atoms: [e, x, y, z, q]
+	np.save(fname+'_atoms.npy',at_array) # at the moment we have not switched npz, I will do it later, maybe #
```
yakutovicha, line 409: I suggest avoiding such kinds of comments. If anything is planned to be done, there should be an issue about that.

```suggestion
	np.save(fname+'_atoms.npy',at_array)
```

ppafm/HighLevel.py
```diff
@@ -146,7 +146,7 @@ def computeLJ( geomFile, speciesFile, save_format=None, computeVpot=False, Fmax=
     FFparams            = PPU.loadSpecies( speciesFile )
     elem_dict           = PPU.getFFdict(FFparams); # print elem_dict
     # --- load atomic geometry
-    atoms,nDim,lvec     = BU.loadGeometry( geomFile, params=PPU.params )
+    atoms,nDim,lvec     = BU.loadGeometry( geomFile, params=PPU.params ) # atoms can be iether [e,x,y,z] or [e,x,y,z,q] for each atoms #
```
yakutovicha, line 149: What is the purpose of the `#` symbol in the terminal position?
```suggestion
    atoms,nDim,lvec     = BU.loadGeometry( geomFile, params=PPU.params ) # atoms can be iether [e,x,y,z] or [e,x,y,z,q] for each atoms.
```

yakutovicha (changes requested):
Thanks, @ondrejkrejci. I have a few suggestions and comments.

ppafm/basUtils.py
ondrejkrejci: I do not think so, as atoms is list of lists [e, x, y, z] or [e, x, y, z, q], where e is interger, and lvec is 4x3 matrix, where I on purpose want to get rid of the first row.

ppafm/HighLevel.py
ondrejkrejci: It is just my way, as I use comments ...

Issue #93, opened 2023-02-28
Adhere to the Python naming convention


yakutovicha:
Here is a long and (I guess) complete guide to Python code style: [PEP8](https://peps.python.org/pep-0008/).

TL;DR

Just do that [1]:

Type | Public | Internal
-- | -- | --
Packages | lower_with_under |  
Modules | lower_with_under | _lower_with_under
Classes | CapWords | _CapWords
Exceptions | CapWords |  
Functions | lower_with_under() | _lower_with_under()
Global/Class Constants | CAPS_WITH_UNDER | _CAPS_WITH_UNDER
Global/Class Variables | lower_with_under | _lower_with_under
Instance Variables | lower_with_under | _lower_with_under
Method Names | lower_with_under() | _lower_with_under()
Function/Method Parameters | lower_with_under |  
Local Variables | lower_with_under

[1] https://namingconvention.org/python/

PR #96, opened 2023-02-28
Add black.

yakutovicha:


yakutovicha:
@NikoOinonen and @ProkopHapala, please look at how `black` would auto-format the files. I configured black to look only at the following files: `generateLJFF.py` and `conv_rho.py`.

Some other changes, like unused imports, I've done manually.

conv_rho.py
```diff
@@ -3,82 +3,110 @@
...
 from optparse import OptionParser
 
-import __main__ as main
-import matplotlib.pyplot as plt
 import numpy as np
 
-import ppafm as PPU
-import ppafm.fieldFFT as fFFT
-import ppafm.GridUtils as GU
+from ppafm import GridUtils
```
NikoOinonen, line 8: Why remove the alias on the import? What logic does it use to decide when an alias is valid or not? Would it replace `import numpy as np` with `import numpy`?

generateLJFF.py
```diff
@@ -3,39 +3,55 @@
...
+        main.__file__
+    )
     from optparse import OptionParser
+
     parser = OptionParser()
-    parser.add_option("-i", "--input",      action="store", type="string",  help="Input file, supported formats are:\n.xyz\n.cube,.xsf")
-    parser.add_option("-f","--data_format", action="store" , type="string", help="Specify the output format of the vector and scalar field. Supported formats are: xsf,npy", default="xsf")
-    parser.add_option("--noPBC",            action="store_false",           help="pbc False", dest="PBC", default=None)
-    parser.add_option("-E", "--energy",     action="store_true",            help="Compue potential energ y(not just Force)", default=False)
-    parser.add_option("--ffModel",          action="store",                 help="kind of potential 'LJ','Morse','vdW' ", default='LJ')
```
NikoOinonen, line 27: Even though these lines are really wide, I would argue that since it's so nicely aligned it reads better than automatically formatted version. You can see all of the arguments at a glance.

generateLJFF.py
```diff
@@ -3,39 +3,55 @@
...
 
-if __name__=="__main__":
-    HELP_MSG="""Use this program in the following way:
+if __name__ == "__main__":
+    HELP_MSG = """Use this program in the following way:
     %s -i <filename>
 
     Supported file fromats are:
        * xyz
-    """ %os.path.basename(main.__file__)
```
NikoOinonen, line 20: Why does it divide this into multiple lines?

conv_rho.py
yakutovicha: That was the change requested by flake that I did manually.

I think I will revert everything I did manually to give one a better overview of what precisely black did.

yakutovicha: ok, I removed this change in [95ff0c7](https://github.com/Probe-Particle/ppafm/pull/96/commits/95ff0c76457fdf14d7de479e0204114becd64254) to create less confusion.

generateLJFF.py
yakutovicha: I don't know this. Probably black considers a docstring as a single line, no idea.

generateLJFF.py
yakutovicha: The only problem is that we are far from being consistent throughout the whole code. Maybe here it reads nice, but in many other places, the number of spaces and alignment is random.

I think consistency is better than having one specific part of the code nicely aligned.

NikoOinonen: I see a couple of places where I think it improves legibility, a couple of places like this where it makes it worse, but mostly just kind of neutral. I think it's quite situation dependent what kind of style is more legible.

Overall, I think these kind of aesthetic changes are not that important and rather the actually important things to make code readable are things like
- Divide the operations in the code into lines or blocks in a sensible way that make the flow of the logic clear.
- Write enough comments to make clear what each part of the code is doing.
- Make the overall structure of the code consistent and make a consistent API. I think switching to PEP8 naming scheme would be a positive change in this regard.

I don't see black helping much with this, but I also don't see value in debating very long about it, so maybe it's better if you just go ahead and implement it.

yakutovicha: > * Divide the operations in the code into lines or blocks in a sensible way that make the flow of the logic clear.

Black doesn't read the logic, it is a code formatter. We have many other pre-commit hooks that help with a better writing style.

> * Make the overall structure of the code consistent

Here black is definitely helpful.

> I think switching to PEP8 naming scheme would be a positive change in this regard.

I opened an issue about this a few days ago: https://github.com/Probe-Particle/ppafm/issues/93

ProkopHapala: Yes, this is exactly example where I worry that black will destroy something which helps me read the code just for the sake of some arbitrary convention. I have really problem reading long code, and I need to organize the code into blocks/tables to help me see the connections (help see the common patterns). Even when I read other people codes I often have to reformat them into such blocks/tables in order to help me see the patterns and understand it.

The importain thing is that this block-structure I have to hand-craft, which is tedious, and Black would destroy it in one go, and there is no way back without doing all that tedious human-labor-intensive re-formating again. Therefore I'm against using black.

I would be quite happy to use some automatic formatar, which can do this for me - i.e. where I can set the rules to do this table-like or block-like alignement automatically.

But also I agree with @NikoOinonen  that I don't want to spend much time on this cosmetic issue (i.e. carefully set the rules for automatic formater).

conv_rho.py
```diff
@@ -13,72 +13,100 @@
...
+    help="prefactor A in formula E = A*Integral( rho_tip^B * rho_sample^B ); NOTE: default A=1 since re-scaling done in relax_scan_PVE.py",
+)
+parser.add_option("-E", "--energy", action="store_true", help="Compue potential energ y(not just Force)", default=False)
+parser.add_option("--saveDebugXsfs", action="store_true", help="save auxuliary xsf files for debugging", default=False)
+parser.add_option(
+    "--densityMayBeNegative",
+    action="store_false",
+    help="input desnity files from DFT may contain negative voxels, lets handle them properly",
+    default=True,
+)
```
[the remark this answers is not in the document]
ProkopHapala: Also consider that it is much easier to understand things which you can see in a whole (does not take too much space). It is difficult to see correlations and patterns over the large distance.

An other factor is that things which are not importaint / not interesting should not take too much space.

ProkopHapala:
I would be probably fine with Black (or other automatic re-formated) if we can set it up so that it will 
1. Not mess up with white-spaces (so that blocks of code which are tediously manually aligned stay aligned)
2. Not mess up with line-breaks (so that calls of functions with many arguments are not expanded into many lines) 

Can we try to switch these two functions of and see the result ?

ProkopHapala:
> Even if it wouldn't contradict PEP8, how would you programmatically make your "... blocks of code which are tediously manually aligned stay aligned...""" discoverable by any automated tool? How is this sort of alignment can be automated at all? And how can you enforce such a standard for the other developers?

I don't care about PEP8. I care about how easy is to see patterns in the code.

What I mean, I think it would be good to use formater with makes sure the *indentation* and using of empty-lines between code blocks (e.g. between functions) is consistent. Or that e.g. comments has consistent format.

But I don't want the formater to mess up with the lines at all (i.e. not tough whitespaces within the line)

>  Your concerns can essentially be rephrased as "I like it this way",

No I think there are some objective criteria of how readeable is the code - i.e. how easy it is to spot patters and perturbances of the pattern.  For example one such objective criteria is that `=` sign in equation is under each other. or that if there is some equivalent name of function or of variable it is alligned under each other, so that you can spot immediately that they are the same. That is the patter. When you make a typo, or when you use different variable at ceraint place, you can spot that perturbance of the pattern much easily. Sticking to PEP8 destroys this.

ProkopHapala:
> > Even if it wouldn't contradict PEP8, how would you programmatically make your "... blocks of code which are tediously manually aligned stay aligned...""" discoverable by any automated tool? How is this sort of alignment can be automated at all? And how can you enforce such a standard for the other developers?
> 
> I don't care about PEP8. I care about how easy is to see patterns in the code.
> 
> What I mean, I think it would be good to use formater with makes sure the **indentation** and using of **empty-lines** between code blocks (e.g. between functions) are consistent. Or that e.g. comments has consistent format.
> 
> But I don't want the formater to mess up with the lines at all (i.e. not touch whitespaces within a single line)
> 
> > Your concerns can essentially be rephrased as "I like it this way",
> 
> No I think there are some objective criteria of how readeable the code is - i.e. how easy it is to spot patters and perturbances of the pattern. For example one such objective criteria is that `=` sign in equation is under each other. or that if there is some equivalent name of function or of variable it is aligned under each other, so that you can spot immediately that they are the same. That is the patter. When you make a typo, or when you use different variable at certaint place, you can spot that perturbance of the pattern much easily. Sticking to PEP8 destroys this.

> that readability comes with consistency, while consistency can only be achieved if automated

Consistency is one of the factors, but not the only and not the most importaint factor for readeability

yakutovicha:
@ProkopHapala, with more people contributing to the code, you will have an equal number of formats that they like. They will have different opinions on what the is "objective criteria of code readability".

What do you want to do there:

1. Let everybody contribute their style and do not enable formatters at all.
2. Review the PRs and make sure the style is consistent with yours.
3. Try to adapt some external formatters to match your style (I have no idea how to do that).
4. Write your own formatter.
5. Something else.

Ultimately, I am fine with any solution here. I just want it to be clearly stated.

NikoOinonen:
> People thought about styling and they came up with quite good guidelines.

I also think that PEP8 is a good guideline, and I mostly stick to that style when writing my own code. But any standard is a compromise, especially for an activity as broad in scope as programming, and there arise situations when a strict standard does not in fact make the code more clear or understandable. 

> But I am afraid none of the formatters will automatically convert this ... into this ...

I wouldn't try to automatically align every equals sign either. I think it's equally nonsensible to insist that it always has to be that way, as it is to insist that it can never be that way.

> If what we want is to disable some parts of the code from re-formatting, this can surely be done

I would be fine with this. Use black, but disable it manually for some specific cases.

ProkopHapala:
@NikoOinonen 
> I would be fine with this. Use black, but disable it manually for some specific cases.

How can we do this? E.g. if it would be as easy as putting some hint in comment like
`#@black disable` before a function or code block, than I'm possibly fine with that.

@yakutovicha 

By default I inclined to this:

`1. Let everybody contribute their style and do not enable formatters at all.`

I was never thinking about code formating as something which should be enforced. I don't want to spend time on bullying people to stick to some superficial formality, causing bad feeling and detering them from creative expression in code. I think the author of the code has best idea how to express his ideas in most suggestive way.

Automatic formater I consider as something which should **help you**, and not **bully you**. People, (especially me) are not machines, so they often do things inconsistently and iregulrarly (e.g. I make a lot of typos). I'm happy if some auto-corrector corrects these typos after me. But I would be very unhappy if some machine forbid me to express my ideas the way I find most clear.

yakutovicha:
> How can we do this? E.g. if it would be as easy as putting some hint in comment like
> `#@black disable` before a function or code block, than I'm possibly fine with that.

Yes, you need to put  `# fmt: off` and `#fmt: on` around the block you don't want to be formatted. Like in the example [here](https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4ARsAnNdAD2IimZxl1N_WlkPinBFoXIfdFTaTVkGVeHShArYj9yPlDvwBA7LhGo8BvRQqDilPtgsfdKl-ha7EFp0Ma6lY_06IceKiVsJ3BpoICJM9wU1VJLD7l3qd5xTmo78LqThf9uibGWcWCD16LBOn0JK8rhhx_Gf2ClySDJtvm7zQJ1Z-Ipmv9D7I_zhjztfi2UTVsJp7917XToHBm2EoNZqyE8homtGskFIiif5EZthHQvvOj8S2gJx8_t_UpWp1ScpIsD_Xq83LX-B956I_EBIeNoGwZZPFC5zAIoMeiaC1jU-sdOHVucLJM_x-jkzMvK8Utdfvp9MMvKyTfb_BZoe0-FAc2ZVlXEpwYgJVAGdCXv3lQT4bpTXyBwDrDVrUeJDivSSwOvT8tlnuMrXoD1Sk2NZB5SHyNmZsfyAEqLALbUnhkX8hbt5U2yNQRDf1LQhuUIOii6k6H9wnDNRnBiQHUfzKfW1CLiThnuVFjlCxQhJ60u67n3EK38XxHkQdOocJXpBNO51E4-f9z2hj0EDTu_ScuqOiC9cI8qJ4grSZIOnnQLv9WPvmCzx5zib3JacesIxMVvZNQiljq_gL7udm1yeXQjENOrBWbfBEkv1P4izWeAysoJgZUhtZFwKFdoCGt2TXe3xQ-wVZFS5KoMPhGFDZGPKzpK15caQOnWobOHLKaL8eFA-qI44qZrMQ7sSLn04bYeenNR2Vxz7hvK0lJhkgKrpVfUnZrtF-e-ubeeUCThWus4jZbKlFBe2Kroz90Elij_UZBMFCcFo0CfIx5mGlrINrTJLhERszRMMDd39XsBDzpZIYV4TcG7HoMS_IF8aMAAAxI-5uTWXbUQAAY8F7QgAAP01Vc6xxGf7AgAAAAAEWVo=).

Is it the compromise solution everyone is happy with?

ProkopHapala:
> Is it the compromise solution everyone is happy with?

Aha, in that case I'm pro. 

Just I think it would take me quite some time to go through the code and put these everywhere I see it important. I don't see it much priority right now.

This is relevant just for python part of the code, right?

yakutovicha:
> > Is it the compromise solution everyone is happy with?
> 
> Aha, in that case I'm pro.

cool 👍 

> Just I think it would take me quite some time to go through the code and put these everywhere I see it important. I don't see it much priority right now.

I can do a bunch of PRs that modify a bunch of related files (not too many) and ask you to mark the changes you don't like. Then I add the `# fmt: off` to `# fmt: on` to those places.

Does this approach work for you?

> This is relevant just for python part of the code, right?

yes

ProkopHapala:
> I can do a bunch of PRs that modify a bunch of related files (not too many) and ask you to mark the changes you don't like. Then I add the # fmt: off to # fmt: on to those places.

OK

NikoOinonen (approved):
:+1:

PR #98, opened 2023-02-28
Update .gitignore.

### LINKED ISSUE #61: Extend .gitignore file for typical things excluded in Python projects.
[Here](https://github.com/github/gitignore/blob/main/Python.gitignore) is an example of such a file. We can just take it as is, or be more mindful and select only the things we need.

yakutovicha:
fixes #61

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,12 @@ dist/
 # Mac
 .DS_Store
 
+# Windows
+*.obj
+*.exp
+*.lib
+*.dll
+
 # Output files
 *.png
 *.np[xz]


.gitignore
```diff
@@ -1,12 +1,24 @@
...
-*.pyd
-*_backup
+*.o
 
-*.npy
-*.npz
-*.xsf
+# Mac
+.DS_Store
+
```
NikoOinonen, line 17: For Windows, lets also add these

```suggestion

# Windows
*.obj
*.exp
*.lib
*.dll

```

NikoOinonen (approved):
Ok, looks good to me.

PR #101, opened 2023-03-02
Fix for deprecated use of numpy data types.

### LINKED ISSUE #76: Using the aliases of builtin types like `numpy.int` is deprecated

NikoOinonen:
Fixes #76 

Replace deprecated numpy data types such as `np.int` with identical built-in data types according to https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations.

### DIFF
--- a/dev/test_PolyCycles2.py
+++ b/dev/test_PolyCycles2.py
@@ -235,10 +235,10 @@ if __name__ == "__main__":
 
     ring_bonds  = ch.findBonds(ring_pos,ring_Rs,fR=1.0)
     ring_neighs = ch.bonds2neighs(ring_bonds,Nring)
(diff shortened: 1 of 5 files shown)


yakutovicha (approved):
Thanks, @NikoOinonen 👍

PR #102, opened 2023-03-02
Install top-level scripts with pip.

### LINKED ISSUE #99: Install scripts with pip
Now that #92 is resolved, we have an actually functional pip package. However, it's still missing the runnable scripts like `generateElFF` and the GUI program, which have to be downloaded separately. Setuptools has a feature for adding these kind of scripts to the pip installation: https://setuptools.pypa.io/en/latest/userguide/entry_point.html.

yakutovicha:
fixes #99


pyproject.toml
```diff
@@ -38,6 +38,13 @@ dev = [
     "pytest-cov",
 ]
 
+[project.scripts]
+ppafm-generate-elff = "ppafm.cli.GenerateELFF:main"
+ppafm-generate-elff-point-charges = "ppafm.cli.GenerateELFF_point_charges:main"
+ppafm-generate-ljff = "ppafm.cli.GenerateLJFF:main"
```
NikoOinonen, line 44: These have typos. Doesn't work on Linux filesystems which are case-sensitive:
```suggestion
ppafm-generate-elff = "ppafm.cli.generateElFF:main"
ppafm-generate-elff-point-charges = "ppafm.cli.generateElFF_point_charges:main"
ppafm-generate-ljff = "ppafm.cli.generateLJFF:main"
```

pyproject.toml
```diff
@@ -38,6 +38,13 @@ dev = [
...
+ppafm-relaxed-scan = "ppafm.cli.relaxed_scan:main"
+ppafm-plot-results = "ppafm.cli.plot_results:main"
+
```
NikoOinonen, line 47: Can we also add the GUI here from `GUI/ppm-gui` and maybe rename it to `ppafm-gui`? The only problem is that it requires the optional `[opencl]` dependencies.

NikoOinonen (changes requested):

pyproject.toml
yakutovicha: done in 5bb1627

PR #103, opened 2023-03-06
IO related code reorganization

### LINKED ISSUE #46: Organize file IO related functions to their own module
According to our previous discussions, it would make sense to group all of the IO related functions to their own module. Currently `basUtils.py` has most of those functions, but it has also some other functions that would better fit into `atomicUtils.py`.

So the proposal would be to rename `basUtils.py` into `io.py` and move the following functions to `atomicUtils.py`: `findBonds`, `findBondsNP`, `findBonds_`, `findBondsSimple`, `multCell`.

Additionally, `GridUtils.py` also has some IO functions, `saveXSF` etc., which could either be moved to the new `io.py` or at least imported there so that all IO function can be accessed in one place.

But we should wait until #24 gets resolved before doing this.

NikoOinonen:
Fixes #46 

Reorganizing code related to file IO so that all IO functions are in one module called `io`:
- Rename `basUtils` to `io`
- Move some functions from `io` to `atomicUtils`
- Move several functions from `GridUtils` to `io`. Basically `GridUtils` now only functions as an interface to the C++ functions.
- Deleted a couple of functions that were not used anywhere and did not seem important.
- Fix all affected imports.


yakutovicha:
@NikoOinonen, please resolve the conflicts. I will then review the PR.

dev/FARFF/test_FARFF_AuxMaps.py
```diff
@@ -10,7 +10,6 @@
 #from ppafm import fieldOCL     as FFcl
 #from ppafm import RelaxOpenCL  as oclr
 #from ppafm import common       as PPU
-#from ppafm import basUtils
 #from ppafm.AFMulatorOCL_Simple  import AFMulator
 #from ppafm.GeneratorOCL_Simple2 import InverseAFMtrainer
 #from ppafm.AuxMap import AuxMaps
```
yakutovicha, line 15: Please remove the unnecessary comments and `sys.path.append`. Same for all the occurrences below.

dev/bakup/GUImorse.py
```diff
@@ -394,8 +390,8 @@ def selectMode(self):
         self.loadInputs_New()
         self.getFF()
 
-        #GU.saveVecFieldXsf( "DEBUG_FFcl", self.FF[:,:,:,:3].astype(np.float32), self.lvec )
-        #GU.saveXSF( "DEBUG_FFcl_z.xsf", self.FF[:,:,:,2].astype(np.float32), self.lvec )
+        #io.saveVecFieldXsf( "DEBUG_FFcl", self.FF[:,:,:,:3].astype(np.float32), self.lvec )
+        #io.saveXSF( "DEBUG_FFcl_z.xsf", self.FF[:,:,:,2].astype(np.float32), self.lvec )
```
yakutovicha, line 394:
```suggestion
```

dev/bakup/conv_rho_mini.py
```diff
@@ -14,20 +14,16 @@
 
 import ppafm as PPU
 import ppafm.fieldFFT as fFFT
-import ppafm.GridUtils as GU
-
-#import GridUtils as GU
-#sys.path.append("/u/25/prokoph1/unix/git/ProbeParticleModel")
-
+from ppafm.io import loadXSF, saveXSF
```
yakutovicha, line 17: I would suggest the following pattern everywhere:
```python
from ppafm import io

io.saveXSF(..)  # everywhere
```

dev/bakup/conv_rho_mini.py
```diff
@@ -53,14 +49,14 @@
...
 
 np.save( "E"+namestr+"_mini.npy", Ecut*(PQ*-1.0) )
 np.savez_compressed( "E"+namestr+"_mini.npz", Ecut*(PQ*-1.0) )
 
-#GU.saveXSF( "FF"+namestr+"_x.xsf", Fx*PQ,       lvec1, head=head1 )
-#GU.saveXSF( "FF"+namestr+"_y.xsf", Fy*PQ,       lvec1, head=head1 )
-#GU.saveXSF( "FF"+namestr+"_z.xsf", Fz*PQ,       lvec1, head=head1 )
+#saveXSF( "FF"+namestr+"_x.xsf", Fx*PQ,       lvec1, head=head1 )
+#saveXSF( "FF"+namestr+"_y.xsf", Fy*PQ,       lvec1, head=head1 )
+#saveXSF( "FF"+namestr+"_z.xsf", Fz*PQ,       lvec1, head=head1 )
```
yakutovicha, line 61: eliminate.

```suggestion
```

dev/bakup/runOpenCLbatch.py
```diff
@@ -159,7 +156,7 @@ def evalFF_LJC( iZPP, xyzs, Zs, qs, poss, typeParams, func_runFF=FFcl.runLJC ):
         FF    = evalFFatoms_LJC( atoms, cLJs, poss, func_runFF=FFcl.runLJC )
         FEin =  FF[:,:,:,:4] + Q*FF[:,:,:,4:]
         Tff = time.clock()-t1ff;
-        #GU.saveXSF( geomFileName+'_Fin_z.xsf',  FEin[:,:,:,2], lvec );
+        # io.saveXSF( geomFileName+'_Fin_z.xsf',  FEin[:,:,:,2], lvec );
```
yakutovicha, line 159: There are many commented lines everywhere in the files you updated. Just remove them, please.

plot_results.py
```diff
@@ -0,0 +1,273 @@
+#!/usr/bin/python -u
```
yakutovicha, line 1: that file should be gone.

yakutovicha (changes requested):
thanks a lot @NikoOinonen. The two main suggestions I have are:

1. Please always use the same coding pattern:
```python
from ppafm import io

io.loadXYZ(..)
```

2. Please remove unnecessary comments.

plot_results.py
NikoOinonen: Deleted.

NikoOinonen:
@yakutovicha Most of the problems are in the scripts inside `dev`, and it looks like most of those scripts are unmaintained and broken. In hindsight, I probably should not have touched them at all, but it was just easier to grep and fix every occurrence of `basUtils` and `GridUtils` than to be selective.

@ProkopHapala Could you check if these scripts are actually needed, and maybe delete or move them to some other branch if they are not important? So that we don't have to spent time modifying them when we do fixes like this.

dev/bakup/runOpenCLbatch.py
ondrejkrejci: I am having a problem as this reorganization is heavily messing with my pull request, which is to be finished this week (if my son is not ill). Maybe I should wait to this one to be pushed and then I will change the NPY part of it?

ondrejkrejci:
I am having a problem as this reorganization is heavily messing with my pull request, which is to be finished this week (if my son is not ill). Maybe I should wait to this one to be pushed and then I will change the NPY part of it?

NikoOinonen:
@ondrejkrejci We can also merge yours first and then pull those fixes here. Whichever way you think is easiest.

ondrejkrejci:
> @ondrejkrejci We can also merge yours first and then pull those fixes here. Whichever way you think is easiest.

If this is quicker than me, go ahead.

yakutovicha:
> I am having a problem as this reorganization is heavily messing with my pull request, which is to be finished this week (if my son is not ill). Maybe I should wait to this one to be pushed and then I will change the NPY part of it?

My 2 cents on this topic.

An open pull request will have more conflicts with time. That is normal because other people are also working on the code. Therefore, resolving merge conflicts is a natural part of collaborative development. An open pull request should never be a reason for others to wait or to stop working.

yakutovicha:
Blocked till https://github.com/Probe-Particle/ppafm/issues/119 is resolved.

yakutovicha:
@NikoOinonen, please merge the `main` branch into it so the tests are passing.

NikoOinonen:
Merged main and all checks are passing.

yakutovicha (approved):
Thanks @NikoOinonen. LGTM.

I think it is a good step forward towards high-quality code 👍

PR #104, opened 2023-03-06
Prepare release 0.2.0a1

yakutovicha:
Preparing the first (hopefully) working release on PyPi.


yakutovicha:
@NikoOinonen, thanks for approving. For your information, I documented the steps to [make a pypi release](https://github.com/Probe-Particle/ppafm/wiki/For-Developers#make-a-new-release):

NikoOinonen:
@yakutovicha Regarding the steps for the release, it seems like using `python setup.py sdist` is deprecated in favor of `python -m build` for some [good reasons](https://stackoverflow.com/a/73607710).

yakutovicha:
> @yakutovicha Regarding the steps for the release, it seems like using `python setup.py sdist` is deprecated in favor of `python -m build` for some [good reasons](https://stackoverflow.com/a/73607710).

Thanks! I think we should then add `build` as a development dependency. I will do that now.

Issue #110, opened 2023-03-09
Problems on compiling ppafm in Win11


lycheehoo:
![Snipaste_2023-03-09_10-38-40](https://user-images.githubusercontent.com/117914902/223901737-99ce7eb1-3776-4eb2-836d-95a8a484bd7d.png)
Dear editor,
I was trying to compile these ppafm codes in Win11 through your installation guide, but when I navigated to cpp folder and compiled with the following command: mingw32-make, it displayed an error as shown below.

NikoOinonen:
Windows support is something that we are figuring out at the moment. There should be pre-built pip wheels available (hopefully) soon. In the meanwhile, the best way would be to follow the instructions at https://github.com/Probe-Particle/ppafm#building-the-code and build it using the Microsoft Visual Studio compiler. I have tested that it works on Win10 but not on Win11.

lycheehoo:
> Windows support is something that we are figuring out at the moment. There should be pre-built pip wheels available (hopefully) soon. In the meanwhile, the best way would be to follow the instructions at https://github.com/Probe-Particle/ppafm#building-the-code and build it using the Microsoft Visual Studio compiler. I have tested that it works on Win10 but not on Win11.

Thanks for your reply. I tried to build these codes using Microsoft Visual Studio compiler, but errors come up as below. I do not know if it depends on Win10 or Win11. Besides, I wonder how it looks like running in GUI, is that contains specific icons and buttons so that we can just click on them to get the simulated AFM images? 


`RuntimeError:` Could not find VsDevCmd.bat

ERROR Backend subprocess exited when trying to invoke `

NikoOinonen:
Did you install the Visual Studio Tools from here: https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2022, and check the option for "Desktop development with C++"?  This should install the `VsDevCmd.bat` that it's asking for. If you did, then it could be that the location for that script is somewhere else on Win11, which needs to be taken into account during the build.

The documentation for the GUI is lacking at the moment. It should work on Windows as well if you can get the pip package installed.

lycheehoo:
Fine. I successfully built these codes on win11 and got ppafm-0.2.0a1.tar.gz and ppafm-0.2.0a1-cp310-cp310-win32.whl, so what is the next process and how to use them? Honestly, I am just a coding beginner, thanks for your patient reply.

NikoOinonen:
[...]

PR #111, opened 2023-03-15
Fixed run.bat for the graphene example.

NikoOinonen:
Fix for the graphene example `run.bat` for Windows. Noticed this while looking into #110.


yakutovicha (approved):
LGTM!

PR #121, opened 2023-03-21
Clean dev

ProkopHapala:
* removed all my non-essential WIP projects (like forcefield relaxation, multipole fitting, density compression, OpengGL visualization and CorrectionLoop) ... they are backed up in branch `prokop-dev`
* moved some test files into `/test` folder (I created new one ... I was thinking you olready created one? Maybe it is in different Branch? There may be conflict?)
* some test files were created by you
* my test file for tip-spline are meant to be checked by human eye (not automatically), therefore I created sub folder `tests/human_eye` ... what you think about ti ?
* there was sub folder `/fitPPM` which is from @yakutovicha (as far as I know). So I wait for him to confirm I should delete it (it is backedup in `prokop_dev` branch) ... or he may decide to do something else with it


yakutovicha:
> * there was sub folder `/fitPPM` which is from @yakutovicha (as far as I know). So I wait for him to confirm I should delete it (it is backedup in `prokop_dev` branch) ... or he may decide to do something else with it

Just keep it there, I will make a follow-up PR to move those.

yakutovicha:
@ProkopHapala, Before we start reviewing, can you update your master branch and merge it into `clean_dev`? It is a bit behind currently.

PR #123, opened 2023-03-22
Revert "Clean dev"

ProkopHapala:
Reverts Probe-Particle/ppafm#121

when merging `main` to `clean_dev` somehow accidently happend that `clean_dev` merged to `main`


ProkopHapala:
I think this "revert" pull-request is useless, I would just cancel/delete it.... the original merge of `clean_dev` into `main` seems fine

NikoOinonen:
Although, something seems to be going funny with the automated tests. It's currently failing on the main branch, but restoring the files in this PR for some reason fixes it.

ProkopHapala:
> Although, something seems to be going funny with the automated tests. It's currently failing on the main branch, but restoring the files in this PR for some reason fixes it.

aha, this is perhaps this problem:
```
ERROR tests/human_eye/test_splines.py - RuntimeError: Could not find compiled extension module in `/home/runner/work/ppafm/ppafm/ppafm/cpp/ProbeParticle_lib.so`. Either check pip installation or enable dynamic compilation by setting the environment variable PPAFM_RECOMPILE=1
ERROR tests/human_eye/test_vdwDaping.py - RuntimeError: Could not find compiled extension module in `/home/runner/work/ppafm/ppafm/ppafm/cpp/ProbeParticle_lib.so`. Either check pip installation or enable dynamic compilation by setting the environment variable PPAFM_RECOMPILE=1
```

If I understand it, the tests which I added into `tests/houman_eye` are now automatically compiled as test (which was not intentional, but probably it is good). Now I just have to change it so in comply with the new system. But I'm not quite sure how the new system (after your modifications) work ?

yakutovicha:
> I mentioned it briefly before, but the editable install does not seem to actually run the build step of the installation process, so the C++ extensions never get compiled. I don't think the installation needs to be editable here, so I would suggest changing it to just
> ```shell
> pip install .[dev] 
> ```
> 
> The other tests don't fail because they don't try to import any of the modules that require the compiled libraries.
> 
> I would say we just scrap this PR and fix the problem with the tests in another PR.
> 
> @yakutovicha

I agree

NikoOinonen:
I opened a new PR https://github.com/Probe-Particle/ppafm/pull/125 for the tests issue and I'll close this one.

PR #125, opened 2023-03-23
Fix CI tests

NikoOinonen:
According to the discussion in https://github.com/Probe-Particle/ppafm/pull/123#issuecomment-1480731196, the CI test workflow is not properly building the package because of an editable install. 
- Change the editable install to a regular pip install.
- Don't run the test for the `tests/human_eye` scripts because at least one of them relies on a correct input argument and those scripts were not meant for automatic testing in the first place.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,4 +24,4 @@ jobs:
                   pip install .[dev]
 
             - name: Run pytest
-              run: pytest tests -v --cov
+              run: pytest tests -v --cov --ignore tests/human_eye
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
 
             - name: Install package and its dependencies
               run: |
-                  pip install -e .[dev]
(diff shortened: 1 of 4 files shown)


yakutovicha (commented):
Would be nice to actually run those tests, but instead of displaying the plots, store them as [GitHub artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts). @NikoOinonen would it be possible to do?

NikoOinonen:
@yakutovicha Well, I modified the scripts to work in the CI workflow and added a step to store the images, and it seems to work. I don't exactly get what the goal of this is, though. The files are there and they can be inspected after a download, but probably no one is going to do that for every push or PR.

yakutovicha:
> I don't exactly get what the goal of this is, though. The files are there and they can be inspected after a download, but probably no one is going to do that for every push or PR.

No need to do that for every PR but in case of issues they are very helpful to understand what happens and when it happens.

yakutovicha (approved):

PR #127, opened 2023-03-28
Completes renaming "basUtils" to "io"

### LINKED ISSUE #126: ppafm-plot-results fails, missing basUtils
The command `ppafm-plot-results` stopped working (at least for me) after some recent updates of the code. The crux of the error message is `ImportError: cannot import name 'basUtils' from 'ppafm'`. In more detail, when I run `ppafm-plot-results -k 0.5 -q -0.05 -a 2.0 --df 2> err.txt` in `ppafm/examples/Graphene` after running all the preceding commands from `ppafm/examples/Graphene/run.sh`, I get the error message which I have uploaded as
[err.txt](https://github.com/Probe-Particle/ppafm/files/11080421/err.txt)
Perhaps this happens because of #103. It seems I can repair this issue simply by replacing the reference to `basUtils` in `ppafm/ppafm/cli/plot_results.py` by `io`. If no one sees any related problem to this issue, I will prepare the corresponding pull request.

mondracek:
Resolves #126.
In `ppafm/cli/plot_results.py`, import `elements` as
```
from ppafm import io, elements
```
rather than
```
from ppafm import basUtils, elements
```
.

### DIFF
--- a/ppafm/cli/plot_results.py
+++ b/ppafm/cli/plot_results.py
@@ -9,7 +9,7 @@ import numpy as np
 
 import ppafm as PPU
 import ppafm.cpp_utils as cpp_utils
-import ppafm.GridUtils as GU
+import ppafm.io
 import ppafm.HighLevel as PPH
 import ppafm.PPPlot as PPPlot
 from ppafm import io, elements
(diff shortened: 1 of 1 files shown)


mondracek:
Oh, now I see `plot_results.py' does not work yet. But I should probably start it as a new issue. The error message with which
```
ppafm-plot-results -k 0.5 -q -0.05 -a 2.0 --df
```
finishes now is
```
error:  (<class 'AttributeError'>, AttributeError("module 'ppafm.GridUtils' has no attribute 'load_scal_field'"), <traceback object at 0x1550ab371200>)
cannot load :  Q-0.05K0.50/OutFz.xsf
 ***** ALL DONE *****
```

PR #130, opened 2023-04-03
Full-density based model in OpenCL

### LINKED ISSUE #35: FDBM on GPU
Just opening this issue to say that I'm working on making an OpenCL implementation of the full density-based model (FDBM) according to https://doi.org/10.1021/acsnano.8b08209 and https://doi.org/10.1021/acs.jcim.1c01323. The CPU-only version already exists via the scripts `conv_rho.py` and `relaxed_scan_PVE.py`. I will make pull request when it's ready.

blocked by #112

NikoOinonen:
Fixes #35 

Implements the full-density based model according to https://pubs.acs.org/doi/10.1021/acs.jcim.1c01323 in OpenCL, with all of the different vdW damping methods consistent with the CPU code.

Additionally, there is some code reorganization in the OpenCL code. Particurlarly in `ocl.field`, there is an added new class called `DataGrid` with methods for easily getting all the required data (electron densities, hartree potential) from disk to the device memory and back, as well as providing some methods to access operations on the OpenCL device.

Also changes the convolution integral into a cross-correlation, which gets rid of an array reversal.


ProkopHapala:
@NikoOinonen Before I do to review, I would like to ask about `ppafm_gui.py` in current `master` branch. When I just run it  without any installation it fails, and before it was working. Do you have idead what is the problem? I guess it is due to some reorganization you did 
(Maybe I can make new issue for this? but anyway)

```
prokophapala@prokophapala-Lenovo-ideapad-Y700-15ISK:~/git/ppafm/ppafm/cli/gui$ ./ppafm_gui.py
Traceback (most recent call last):
  File "/home/prokophapala/git/ppafm/ppafm/cli/gui/./ppafm_gui.py", line 16, in <module>
    import ppafm.common as PPU
ModuleNotFoundError: No module named 'ppafm'

```

.github/workflows/ci.yml
```diff
@@ -21,7 +21,8 @@ jobs:
 
             - name: Install package and its dependencies
               run: |
-                  pip install .[dev]
+                  pip install .[dev,opencl]
+                  sudo apt install pocl-opencl-icd
```
ProkopHapala, line 25: this is required or optional dependence? 
(sorry, I don't understand these things)

ppafm/cpp/Forces.h
```diff
@@ -181,23 +181,23 @@ inline double R4_func(double r2,double &df){
...
+        double e = 1 - r2;
         df     =  -4*e;
         return     e*e;
     }
 }
 
 inline double invR4_func(double r2,double &df){
     //printf( "invR4_func() \n");
-    double invR2 = 1/r2;
+    double invR2 = 1/(r2 + R2SAFE);
```
ProkopHapala, line 192: why you remove R2SAFE ? For precission ?
I guess in practice it does not matter since the atoms never get too close, but anyway ...

ppafm/ocl/cl/FF.cl
```diff
@@ -619,18 +748,18 @@ __kernel void gradPotential(
     int jn = j < nGrid.y - 1 ? ind + nGrid.z : ind - (nGrid.y - 1) * nGrid.z;
```
ProkopHapala, line 748: this is idex wraping due to PBC? maybe put there coment that this grad works in PBC ?

ppafm/ocl/cl/FF.cl
```diff
@@ -592,13 +720,14 @@ __kernel void evalLJC_Hartree(
...
 
-// Obtain electric field as the negative gradient of Hartree potential via centered difference
-__kernel void gradPotential(
-    __global float  *pot,   // Input Hartree potential
-    __global float4 *field, // Output electric field
-    int4            nGrid,  // Size of Hartree potential grid (x, y, z, _)
-    float4          step,   // Step size of grid (x, y, z, _)
-    int             C_order // If non-zeo, values are saved in C order, otherwise Fortran order
+// Compute the gradient of a scalar field via centered difference
+__kernel void grad(
```
ProkopHapala, line 724: this is actually very usefull to have in general

ppafm/ocl/cl/FF.cl
```diff
@@ -719,6 +848,31 @@ __kernel void interp_at(
 
 }
 
+// Raise all array elements to the same power. Negative values are set to zero.
```
ProkopHapala, line 851: As I understand you do the formula `E = A  ( rho1 * rho2  )^b` step-by-step by calling new kernell for each operation. That is probably great for debugging, but I guess for performance it would be better to do it at once just at the end of the convolution (` rho1 * rho2` ) ? (but perhaps it is just very minor overhead)

ppafm/ocl/field.py
```diff
@@ -37,89 +41,356 @@ def makeDivisibleUp( num, divisor ):
...
+            grid_out = array_type(array_out, lvec=self.lvec, shape=self.shape, ctx=self.ctx)
+        n = np.int32(array_in1.size / 4)
+        scale = np.float32(scale)
+        global_size = [int(np.ceil(n / local_size[0]) * local_size[0])]
+        cl_program.addMult(queue, global_size, local_size, array_in1, array_in2, array_out, n, scale)
+        return grid_out
+
+    def power_positive(self, p=1.2, in_place=True, local_size=(32,), queue=None):
+        '''
+        Raise every positive element in the grid into a power. Negative values are set to zero.
```
ProkopHapala, line 220: `negative values are set to zero` seems a bit suspicious. I guess you assume grid contains electron density, which should be possitive everywhere, but due to numerical errors it is sometimes somewhere negative.

I see possible problem that if you set these negative values to zero, you change total amout of chrge in the system. I guess it is very small amout? But maybe it woul be better to renormalize it or something? I know that in some DFT codes using grids, there is renormalization for similar reason.

ppafm/ocl/field.py
```diff
@@ -37,89 +41,356 @@ def makeDivisibleUp( num, divisor ):
...
-    FF *= -1
-    pot = HartreePotential(FF, lvec)
-    xyzs = np.stack([x, y, z], axis=1)
-
-    return pot, xyzs, Zs
-
 # ========= classes
 
-class HartreePotential:
+class DataGrid:
```
ProkopHapala, line 46: Cool, this seems like great way how to organize ti

ppafm/ocl/field.py
```diff
@@ -850,101 +1012,171 @@ def calc_force_fft(self, FE=None, rot=np.eye(3), rot_center=np.zeros(3), local_s
...
+        # Add the two energy contributions together
+        E = E_es.add_mult(E_pauli, scale=A, in_place=True, local_size=local_size, queue=self.queue)
+        E_pauli.release()
+
+        if bRuntime:
+            self.queue.finish()
+            print("runtime(ForceField_LJC.calc_force_fdbm.add_mult) [s]: ", time.perf_counter() - t0)
+
+        # Take gradient to get the force field
+        E.grad(scale=[-1.0, -1.0, -1.0, 1.0], array_out=self.cl_FE, order='F',
```
ProkopHapala, line 1115: aha, so you do not compute directly forces by FFT? Perhaps it is fine. I actually don't know why I did all the hassle to compute forces by FFT, but as I remember I had some problem with numberical noise from finite-differences which was suppressed by using FFT direcly

ppafm/ocl/cl/FF.cl
ProkopHapala: Aha sorry, I did not think about it deeply enouh. I is necessary to do the power `rho^b` first, and then convolution. Than it cannot be put int one kernel perhaps (?).

ProkopHapala (approved):
Great stuff !!! I'm really impressed by how nicely you organized the grids with instance on both GPU and CPU, and that now you have basically something like numpy runnung on GPU where you can easily express arbitery mathematical formula with the grids. 

I did not tested to run it, I relay you did a lot of test during the development. I would like to see some results, especially how it works with `ppafm_gui` :-)

.github/workflows/ci.yml
NikoOinonen: This is just to modify how the automatic tests run here on Github after a push. It adds the necessary dependencies for testing the OpenCL code.

ppafm/cpp/Forces.h
NikoOinonen: No, I added it. It was missing and was causing some instability when I was testing the OpenCL implementation against yours.

ppafm/ocl/cl/FF.cl
NikoOinonen: Yes, I think you have to do it this way around. Although, I'm not sure what the Fourier transform does to it. Maybe there is some more clever way to do it.

ppafm/ocl/field.py
NikoOinonen: Hmm, I was thinking that I just did it the same you did in `conv_rho.py`, but now I realize there is also a rescaling happening there at https://github.com/Probe-Particle/ppafm/blob/710d046686a2e9f0109556edc681db032f8f4131/ppafm/cli/conv_rho.py#L21-L24
This is more challenging to implement in a single kernel, since you would have to do two sums over the whole array. I think I will just open another issue for this.

ppafm/ocl/field.py
NikoOinonen: In my testing there is a little bit of difference between the FFT derivative and the finite difference derivative, but the final force field is almost identical.  I would maybe like to test at some point if using a 5-point stencil instead of 3 makes a difference.

ppafm/ocl/cl/FF.cl
NikoOinonen: Yes, it's for PBC. I added comments both to the OCL kernel and the Python wrapper method.

NikoOinonen:
@ProkopHapala I did not implement this in the GUI yet. I want to do it at some point, but I will have to think how to best interface it with all of the required input files and parameters etc.

I looked at the differences between the OCL and CPU calculated force fields, and they are almost identical. The final scans look a bit different, but I think this is actually due to some difference in the relaxation parameters (which is something that should be investigated).

CPU:
df_017

OCL:
df_017

The different Tip_z is due to different calculation of the distance (which I actually think is bugged in `plot_results.py`)

ppafm/ocl/field.py
NikoOinonen: Added new issue #136.

ProkopHapala:
> I looked at the differences between the OCL and CPU calculated force fields, and they are almost identical. The final scans look a bit different, but I think this is actually due to some difference in the relaxation parameters

Yes, I also think so

PR #131, opened 2023-04-04
Remove remaining sys.path manipulation.

NikoOinonen:
There were some sys.path manipulation still going on in some of the scripts, which is not needed anymore. This removes those.


yakutovicha (approved):
LGTM!

PR #132, opened 2023-04-04
Rename basUtils to io also in sphinx docs.

NikoOinonen:
Two previous PRs #103 and #127 renamed the `basUtils` module to `io`. This changes the sphinx docs to also reflect that change.


yakutovicha (approved):
LGTM!

PR #139, opened 2023-04-06
Parameter saving/loading from params.ini for AFMulator and the GUI

### LINKED ISSUE #80: Some way of saving/loading state

NikoOinonen:
Fixes #80 

Adds methods for constructing or loading an AFMulator instance from a params.ini file and saving the current settings to a params.ini file. Also adds the corresponding bindings to the GUI script and buttons to the GUI for accessing this feature.

This should help in bridging the gap between the usage of the CPU scripts and the OpenCL API/GUI.

This required adding the entries `FFgrid{0,A,B,C}` to the parameter dict to denote the force field grid origin and lattice vectors, because the AFMulator handles the force field lattice separate from the sample lattice vectors. I hope that this does not cause any confusion since these are not used by the CPU scripts.


ppafm/cli/gui/ppafm_gui.py
```diff
@@ -246,11 +246,16 @@ def updateRotation(self):
...
         self.update()
 
-    def updateParams(self):
+    def updateParams(self, preset_none=True):
         '''Get parameter values from input fields and update'''
 
         if self.xyzs is None: return
 
+        if preset_none:
+            self.slPreset.blockSignals(True)
```
ProkopHapala, line 255: What is this? I undrestand this is some locking / synchronization due to parallel access, but woudl be nice to explain it? (e.g. docstring?) It is something general, or your own solution?

ppafm/ocl/AFMulator.py
```diff
@@ -392,6 +392,87 @@ def evalAFM(self, X=None):
...
+            f.write(f'stiffness {k[0]} {k[1]} {k[3]}\n')
+            f.write(f'r0Probe {self.tipR0[0]} {self.tipR0[1]} {self.tipR0[2]}\n')
+            f.write(f'PBC {(np.array(self.npbc) > 0).any()}\n')
+            f.write(f'nPBC {self.npbc[0]} {self.npbc[1]} {self.npbc[2]}\n')
+            if self.sample_lvec is not None:
+                f.write(f'gridA {self.sample_lvec[0, 0]} {self.sample_lvec[0, 1]} {self.sample_lvec[0, 2]}\n')
+                f.write(f'gridB {self.sample_lvec[1, 0]} {self.sample_lvec[1, 1]} {self.sample_lvec[1, 2]}\n')
+                f.write(f'gridC {self.sample_lvec[2, 0]} {self.sample_lvec[2, 1]} {self.sample_lvec[2, 2]}\n')
+            f.write(f'FFgrid0 {self.lvec[0, 0]} {self.lvec[0, 1]} {self.lvec[0, 2]}\n')
+            f.write(f'FFgridA {self.lvec[1, 0]} {self.lvec[1, 1]} {self.lvec[1, 2]}\n')
```
ProkopHapala, line 463: Wow, I actually did not know it is possible to `sprintf` like this in python :)

ppafm/ocl/AFMulator.py
```diff
@@ -450,6 +531,64 @@ def plot_images(self, X, outdir='afm_images', prefix='df'):
         plotImages(os.path.join(outdir, prefix), X, slices = list(range(0, len(X))),
             zs=zTips, extent=extent, cmap=PPU.params['colorscale'])
 
+def _get_params(file_path):
+    '''Get AFMulator arguments from a params.ini file.'''
+    PPU.loadParams(file_path)
+    lvec = np.array([
+        PPU.params['FFgrid0'],
```
ProkopHapala, line 538: loolin on `sample_lvec` and `lvec` side-by-side I'm thinking maybe it would be better to keep origin `FFgrid0` as the last not the first. It would allow to always read first 3 vectors as a,b,c latticle vectros, oblivious if there is 4th vector. 
....
But that is topic for an other issue, not for this PR.

ProkopHapala (approved):
Great. I think this is very useful functionality, and I always learn form you how to organize python code nicely. I cannot judge if there are any bugs from looking at the code, but you wrote the test which pass so it is fine I guess. I approve it. Just please explain me how it works with those locks.

```
self.slPreset.blockSignals(True)
self.slPreset.setCurrentIndex(-1)
self.slPreset.blockSignals(False)
```

ppafm/cli/gui/ppafm_gui.py
NikoOinonen: I'm not that well-versed with pyQt, so there could be some more elegant solution. The reason why those blocks are there is because otherwise it would execute the function bound to the widget (selection box here).

Whenever we connect a function to the widget like 
```python
self.slPreset.currentIndexChanged.connect(self.applyPreset)
```
then all changes to the widget will cause the function to be called, including changes made programmatically. In this case I just wanted to change the display value of the widget without calling the function, so I temporarily block the signals and then enable them again after the change has been made.

ppafm/ocl/AFMulator.py
NikoOinonen: They are called f-strings. They were introduced in Python 3.6. It looks pretty nice and supposedly also executes faster than the old string formatting style: https://realpython.com/python-f-strings/.

ppafm/ocl/AFMulator.py
NikoOinonen: That could be useful, but also would affect quite a few place, so you would have to go carefully through all of the logic that uses `lvec`.

ppafm/cli/gui/ppafm_gui.py
ProkopHapala: please try to mention it briefly in the code. This is exactly one of those things which not obvious because it is `non-local` and therefore difficult to figure out

NikoOinonen: I created another PR for this: #140.

PR #145, opened 2023-04-13
Fix coordinate offset in OpenCL force field interpolation.

### LINKED ISSUE #142: Differences in probe relaxation between the C++ and OpenCL code
I was noticing that the scans look slightly different between the C++ and OpenCL versions of the code. I thought before that this is just a difference from using some faster approximations for the force field calculation in the OpenCL version, but the difference is also there when the force fields are not any different.

I did some testing with the graphene example where I loaded the force field calculated by the regular run script into the OpenCL force field buffer and then doing just the relaxation on the GPU. I get images that look something like this:
df_comp
They look very close, but not identical.

I noticed that the default time step parameters are different. In the C++ code it's hard-coded to `0.1` here: https://github.com/Probe-Particle/ppafm/blob/47cb4684bdbf6fab907109d3240764a62fe5ecaa/ppafm/cpp/ProbeParticle.cpp#L72
whereas in the OpenCL code it's set to `0.5` here: https://github.com/Probe-Particle/ppafm/blob/47cb4684bdbf6fab907109d3240764a62fe5ecaa/ppafm/ocl/AFMulator.py#L65

I tried changing the time step in the OpenCL to test if it would converge to the C++ result. This is the result for a line scan over the horizontal direction in the middle:
df
Far from converging, there is actually a significant jump when the time step is the same `0.1`. There might also be some different prefactor, since the C++ values are consistently smaller.

@ProkopHapala Do you know if there is some other difference in the relaxation algorithms between the two versions the way you implemented them? I don't expect them to return the exact same answer, but it would be nice to know where the difference comes from.

NikoOinonen:
Fixes #142 

The OpenCL built-in `read_imagef` function uses a convention where the coordinates are shifted by half a grid step. This offset was missing, which caused the force to be calculated wrong in the probe particle relaxation algorithm.

### DIFF
--- a/ppafm/ocl/cl/relax.cl
+++ b/ppafm/ocl/cl/relax.cl
@@ -87,7 +87,7 @@ void move_LeapFrog( float3 f, float3 p, float3 v, float2 RP ){
 
 
 //#define N_RELAX_STEP_MAX  64
-#define N_RELAX_STEP_MAX  1000
+#define N_RELAX_STEP_MAX  128
 //#define N_RELAX_STEP_MAX  256
 #define F2CONV  1e-8f
 //#define F2CONV  1e-10f
--- a/ppafm/ocl/cl/relax.cl
+++ b/ppafm/ocl/cl/relax.cl
@@ -62,9 +62,14 @@ float4 read_imagef_trilin( __read_only image3d_t imgIn, float4 coord ){
 
 
 float4 interpFE( float3 pos, float3 dinvA, float3 dinvB, float3 dinvC, __read_only image3d_t imgIn ){
-    const float4 coord = (float4)( dot(pos,dinvA),dot(pos,dinvB),dot(pos,dinvC), 0.0f );
+    float4 offset = (float4)(
+        0.5f / (float) get_image_width(imgIn),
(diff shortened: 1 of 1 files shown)


NikoOinonen:
I also played around with the max relax steps and the relax time step, and in the end I think the current settings are good. The result with 128 max steps and 0.5 time step (current OpenCL settings) is almost identical with 1000 max steps and 0.1 time step (which the CPU code uses), but it's ~4x faster.

ProkopHapala (approved):
Great! :)

NikoOinonen:
> Yes, I remember I was doing similar rests as some point, but forgot the results. Maybe would be usefull to make short summary on wiki ?

Not sure where this would fit it. I think we don't have anything about the OpenCL code in the wiki in at the moment.

ProkopHapala:
> > Yes, I remember I was doing similar rests as some point, but forgot the results. Maybe would be usefull to make short summary on wiki ?
> 
> Not sure where this would fit it. I think we don't have anything about the OpenCL code in the wiki in at the moment.

I would not think to hard about organizing wiki pages. I would just make a page, note what you have, and it is rather easy to reshuffle the text later.

Issue #144, opened 2023-04-12
Try tri-cubic interpolation of just Energy rather than tri-linear interpolation of Force


ProkopHapala:
I think it would be nice to try cubic interpolation. 
* On one hand it would require 64 (4x4x4) reads from memory rather than 8 (2x2x2) of tri-linear
*  but the advantage is that it would not need to calculate and save forces because C1 smooth forces can be calculated analytically from the cubic spline derivative. Therefore the arrays would be smaller (just `float` rather than `float4`). 
* Further reduction of data-size will be probably achieved due to using less dense sampling step because cubic spline provides sufficient accuracy on larger range than linear spline. 

* Maybe the cost of global memory read can be reduced if we read together all the values which are in the same cashe-line. Using buffer it would look something like this:

```
int yz = nx*(ix+ny*iz);
float4 vals = (float4){ grid[ix+yz], grid[ix+1+yz],grid[ix+2+yz],grid[ix+3+yz] };
```

* but I'm not sure if it is really better to use buffer rather than image, because I have feeling texture reads are faster than buffer reads.

* I was asking about this topic on stackoverflow, but did not get any uselful answer yet.
https://stackoverflow.com/questions/75966359/efficient-tricubic-interpolation-of-3d-image-on-gpu-using-opencl

_Originally posted by @ProkopHapala in https://github.com/Probe-Particle/ppafm/issues/142#issuecomment-1505436857_

ProkopHapala:
I skethed the tricubic interpolation krenell here (it is using image `read_imagef()`, maybe would be better version with buffer?)

```
float8 hspline_basis( float x ){
    // hermite spline with derivatives https://en.wikipedia.org/wiki/Cubic_Hermite_spline
    float x2   = x*x;
    float K    =  x2*(x - 1);
    float d0   =  K - x2 + x;   //      x3 -   x2  
    float d1   =  K         ;   //      x3 - 2*x2 + x
    return (float8){
    // ------ values
                     d0*-0.5,
      2*K - x2 + 1 - d1*-0.5,   //    2*x3 - 3*x2 + 1
     -2*K + x2     + d0* 0.5,   //   -2*x3 + 3*x2
                     d0*0.5,
    // ----- derivatives
                     d0*-0.5,
         2*K - d1*-0.5,   //    6*x2 - 6*x
        -2*K + d0* 0.5,   //   -6*x2 + 6*x
               d0*0.5                
                     };
}

[...]

PR #147, opened 2023-04-17
Implement tip density resampling.

### LINKED ISSUE #138: Tip electron density resampling in FDBM in OpenCL
The AFMulator interface to the OpenCL code by default resamples the Hartree potential grid to the chosen force field grid in order to minimize the required computations and memory (getting rid of large vacuum, etc.). However, the full-density based model implemented in #130 does not support this yet and throws an error if all of the grid sizes don't match exactly. Resampling for the tip electron density is more tricky, because the grid needs to be cropped/padded in the middle to match the physical size of the force field grid in addition to interpolation.

NikoOinonen:
Fixes #138

Implements tip density resampling for use with the density overlap model. This enables the use of an arbitrary force field grid, and the grids for the sample density and hartree potentials do not need to match the tip density grid either. So the tip density has to be only calculated once, and can then be easily reused.

Due to reduced force field grid size, this results in quite a significant reduction in the run time for the pyridine example, ~7x on my laptop, and also reduces required device memory.

### DIFF
--- a/ppafm/ocl/cl/FF.cl
+++ b/ppafm/ocl/cl/FF.cl
@@ -851,6 +851,54 @@ __kernel void interp_at(
 
 }
 
+// Interpolate a tip density array onto a new grid. The tip is assumed to be centered
+// on the origin of the input grid, so any resizing of the grid happens in the middle.
+__kernel void interp_tip_at(
+    __global float *in,     // Input array
+    __global float *out,    // Output array
+    int4   nGrid_in,        // Size of input array
+    float4 T_A,             // Rows of the transformation matrix for input array lattice coordinates
+    float4 T_B,
+    float4 T_C,
+    float4 vec_in_inv_A,        // Rows of the inverse of the input array lattice vector matrix
+    float4 vec_in_inv_B,
+    float4 vec_in_inv_C,
+    int4   nGrid_out,       // Size of target grid
+    float4 step_out_A,      // Real-space step sizes of output array lattice vectors
+    float4 step_out_B,
+    float4 step_out_C
+){
+
+    int ind = get_global_id(0);
+    if (ind >= nGrid_out.x * nGrid_out.y * nGrid_out.z) return;
+
+    // Convert linear index to x,y,z indices of output array
+    int i = ind / (nGrid_out.y * nGrid_out.z);
+    int j = (ind / nGrid_out.z) % nGrid_out.y;
+    int k = ind % nGrid_out.z;
+
+    // Past half-way point in the target grid we need to wrap around
+    if (i > nGrid_out.x / 2) i -= nGrid_out.x;
+    if (j > nGrid_out.y / 2) j -= nGrid_out.y;
+    if (k > nGrid_out.z / 2) k -= nGrid_out.z;
+
+    // Calculate position in target grid
+    float4 pos = i * step_out_A + j * step_out_B + k * step_out_C;
+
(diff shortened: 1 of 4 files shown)


NikoOinonen:
The final images are not quite identical with the resampling, so the interpolation does introduce some error. Probably using cubic instead of linear interpolation would be useful here as well (#144).

ppafm/ocl/field.py
```diff
@@ -409,6 +409,53 @@ def subCores(self, xyzs, Zs, Rcore=0.7, valElDict=None):
...
+        if array_out is None:
+            array_out = cl.Buffer(self.ctx, cl.mem_flags.READ_WRITE, size=size_new)
+        else:
+            assert array_out.size == size_new, f'array size does not match ({array_out.size} != {size_new})'
+        array_out = TipDensity(array_out, lvec_new, shape_new, self.ctx)
+
+        global_size = [int(np.ceil(np.prod(shape_new[:3]) / local_size[0]) * local_size[0])]
+        T = np.concatenate([np.linalg.inv(self.step).T.copy(), np.zeros((3, 1))], axis=1, dtype=np.float32)
+        lvec_in_inv = np.concatenate([np.linalg.inv(self.lvec[1:]), np.zeros((3, 1))], axis=1, dtype=np.float32)
+        dlvec_out = np.concatenate([array_out.step, np.zeros((3, 1))], axis=1, dtype=np.float32)
```
ProkopHapala, line 445: I would prefer formating like this (Not just to piss-off Sasha :) )
```
        global_size = [int(np.ceil(np.prod(shape_new[:3]) / local_size[0]) * local_size[0])]
        T           = np.concatenate([np.linalg.inv(self.step).T.copy(), np.zeros((3, 1))], axis=1, dtype=np.float32)
        lvec_in_inv = np.concatenate([np.linalg.inv(self.lvec[1:]),      np.zeros((3, 1))], axis=1, dtype=np.float32)
        dlvec_out   = np.concatenate([array_out.step,                    np.zeros((3, 1))], axis=1, dtype=np.float32)
```

ppafm/ocl/cl/FF.cl
```diff
@@ -851,6 +851,54 @@ __kernel void interp_at(
...
+
+    // Calculate position in target grid
+    float4 pos = i * step_out_A + j * step_out_B + k * step_out_C;
+
+    // Figure out fractional coordinates in the input grid
+    float a_frac = dot(vec_in_inv_A, pos);
+    float b_frac = dot(vec_in_inv_B, pos);
+    float c_frac = dot(vec_in_inv_C, pos);
+
+    // If we go beyond half-way in the input lattice, we pad with zeros
```
ProkopHapala, line 893: Why not to make it simply periodic ?

ProkopHapala (approved):

ppafm/ocl/cl/FF.cl
NikoOinonen: It is periodic, but the period goes from -half to +half of the unit cell so that the padding happens in the middle of the cell and the corners remain fixed. It's kind of hard to see without drawing a picture, but I'm quite sure that this is the correct way to do it.

ppafm/ocl/field.py
NikoOinonen: Well, I don't have a clear preference either way, but I would rather not spend time worrying too much about the formatting for these kind of almost boilerplate type lines.

PR #166, opened 2023-05-26
Implemented DFT-D3 in the CPU code.

### LINKED ISSUE #165: Implement DFT-D3 in the CPU code as well
The DFT-D3 calculation was implemented in OpenCL in #163. This should be implemented in the CPU version as well for feature parity.

NikoOinonen:
Fixes #165 

I went ahead and also implemented the DFT-D3 calculation in the CPU code now that it's still fresh in my mind. I added a new script `generateDFTD3.py` for accessing this feature. It replaces the old LJ script in the pyridine example run script. Also added a test to make sure that the CPU and GPU versions return the same result for the force field.

### DIFF
--- a/ppafm/cpp/Forces.h
+++ b/ppafm/cpp/Forces.h
@@ -14,6 +14,7 @@
 #define RSAFE   1.0e-4f
 #define R2SAFE  1.0e-8f
 #define F2MAX   10.0f
+#define R2_D3_CUTOFF 400.0f
 
 void sum(int n, Vec3d* ps, Vec3d& psum){ for(int i=0;i<n;i++){ psum.add(ps[i]); } };
 
@@ -201,6 +202,28 @@ inline double invR8_func(double r2,double &df){
     double invR8 = invR2*invR2; invR8*=invR8;
     df           = -8*invR8*invR2;
     return            invR8;
+}
+
+// DFT-D3 force between two atoms
+inline double addAtomDFTD3(const Vec3d& dR, Vec3d& fout, double c6, double c8, double R0_6, double R0_8) {
+
+    double r2 = dR.norm2();
+    if (r2 > R2_D3_CUTOFF) return 0;
+    double r4 = r2 * r2;
+    double r6 = r4 * r2;
+    double r8 = r6 * r2;
+
+    double d6 = 1.0f / (r6 + R0_6);
+    double d8 = 1.0f / (r8 + R0_8);
+    double E6 = -c6 * d6;
+    double E8 = -c8 * d8;
+    double F6 = E6 * d6 * 6 * r4;
+    double F8 = E8 * d8 * 8 * r6;
+
+    double E = E6 + E8;
+    fout.add_mul(dR, -(F6 + F8));
+
+    return E;
 }
 
 // Morse force between two atoms a,b separated by vector dR = Ra - Rb
--- a/ppafm/cpp/ProbeParticle.cpp
+++ b/ppafm/cpp/ProbeParticle.cpp
@@ -23,6 +23,9 @@
 
 const double const_eVA_SI = 16.0217662;
 
+#define MAX_REF_CN 5
+#define MAX_D3_ELEM 94
+
 // ================= GLOBAL VARIABLES
 
 GridShape gridShape;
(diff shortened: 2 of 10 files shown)


NikoOinonen:
I also did some testing of the coefficient calculation between the CPU and the GPU versions. For small systems (<~100 atoms), using the CPU calculation for the coefficients before the force field calculation on the GPU is a tiny bit faster (~1ms), just because it takes some time to transfer the reference coefficient arrays to the GPU memory. But for larger systems it's clearly faster to do the whole thing on the GPU, so I left it the way it was.

ppafm/cpp/ProbeParticle.cpp
```diff
@@ -341,6 +345,118 @@ DLLEXPORT void getInPoints_LJ( int npoints, double * points_, double * FEs, int
...
+
+        // Compute the coordination number for this atom
+        double cn = 0;
+        for (int j = 0; j < natoms; j++) {
+            if (j == ia) continue; // No self-interaction for coordination number
+            double d = (Ratoms[j] - pos).norm();
+            double r = r_cov[elems[j] - 1] + r_cov_elem;
+            cn += 1.0f / (1.0f + exp(k12 * r / d + k1));
+        }
+
```
ProkopHapala, line 396: Do I understand here that coordication number is calculated from how many other atoms overlap with this atom (using gaussian kernell) ?

ppafm/cpp/ProbeParticle.cpp
```diff
@@ -341,6 +345,118 @@ DLLEXPORT void getInPoints_LJ( int npoints, double * points_, double * FEs, int
...
+                double L_ab = L_a * L_pp[b];
+                norm += L_ab;
+                L[a * MAX_REF_CN + b] = L_ab;
+            }
+        }
+        int max_ref_a = a;
+
+        // If the coordination number is so high that the gaussian weights are all zero,
+        // then we put all of the weight on the highest reference coordination number.
+        if (norm == 0) {
```
ProkopHapala, line 416: Exactly zero ? I would think there should be some trashold as gaussians are never exactly zero. But probably I don't understand the algorithm?

ppafm/cpp/ProbeParticle.cpp
```diff
@@ -341,6 +345,118 @@ DLLEXPORT void getInPoints_LJ( int npoints, double * points_, double * FEs, int
...
+        int n_zw = MAX_REF_CN * MAX_REF_CN;
+        int n_yzw = MAX_D3_ELEM * n_zw;
+        int pair_ind = elem_ind * n_yzw + pp_ind * n_zw;  // ref_c6 shape = (MAX_D3_ELEM, MAX_D3_ELEM, MAX_REF_CN, MAX_REF_CN)
+        double c6 = 0;
+        for (int a = 0; a < max_ref_a; a++) {
+            int a_ind = a * MAX_REF_CN;
+            for (int b = 0; b < max_ref_pp; b++) {
+                int L_ind = a_ind + b;
+                int c6_ind = pair_ind + L_ind;
+                c6 += L[L_ind] * ref_c6[c6_ind];
```
ProkopHapala, line 435: Hmm, I'm perhaps not able to follow the algorithm, (I should read the article probably) ...

But the basic idea is that the more neighbors the atom have, the more polarizable it is, therefore the higher it the C6 coef? (Or am I completely off?) 

Does it consider shomehow that conjugated sp2 systems are much more delocalized than alifactic sp3 systems?

tests/test_vdw.py
```diff
@@ -56,3 +57,59 @@ def test_vdw():
...
+    PPU.params['gridB'] = lvec[2]
+    PPU.params['gridC'] = lvec[3]
+    FF_cpp, E_cpp = PPH.prepareArrays(None, True)
+    PPC.setFF_shape(np.shape(FF_cpp), lvec)
+    PPC.getDFTD3FF(xyzs, coeffs_cpp)
+
+    FF_cpp = FF_cpp.transpose((2, 1, 0, 3))
+    E_cpp = E_cpp.transpose((2, 1, 0))
+
+    assert np.allclose(coeffs_ocl, coeffs_cpp)
```
ProkopHapala, line 113: very nice, so it test both CPU and GPU. I'm thinking maybe it would be good to have example using just CPU (if somebody does not have OpenCL installed), but that can be done anytime in future

ppafm/cpp/Forces.h
```diff
@@ -203,6 +204,28 @@ inline double invR8_func(double r2,double &df){
     return            invR8;
 }
 
+// DFT-D3 force between two atoms
+inline double addAtomDFTD3(const Vec3d& dR, Vec3d& fout, double c6, double c8, double R0_6, double R0_8) {
+
+    double r2 = dR.norm2();
```
ProkopHapala, line 210: maybe I would add some tiny safety value (like `double r2 = dR.norm2() + 1.e-8;`) just in case when the two atoms accidently overlap. But perhaps this does not solve much.

ProkopHapala (approved):
Perfect. Thanks a lot Niko!

ppafm/cpp/ProbeParticle.cpp
NikoOinonen: Yes, it's a measure of how many atoms are nearby. It's not really Gaussian. It's a smooth step function with a radius proportional to the covalent radius of the two atoms, summed over all other atoms.

ppafm/cpp/ProbeParticle.cpp
NikoOinonen: Yes, in principle a Gaussian is never zero, but due to finite precision, if the distance is big enough, then it will round to exactly zero, and you end up dividing by zero later, so I put this step in. It might not be a problem with double precision (didn't test), but it was definitely a problem with single precision in OpenCL, so I put it here as well.

ppafm/cpp/ProbeParticle.cpp
NikoOinonen: There is a big table of reference values for the C6 coefficients for different coordination numbers. For every pair of elements there is a 5x5 matrix of values. For example, for H-C pair, the matrix is
```
[[ 7.2540984,  6.807746 ,  5.628885 ,  5.270787 ,  4.401504 ],
 [11.511539 , 10.789845 ,  8.820874 ,  8.245396 ,  6.769922 ],
 [ 0.       ,  0.       ,  0.       ,  0.       ,  0.       ],
 [ 0.       ,  0.       ,  0.       ,  0.       ,  0.       ],
 [ 0.       ,  0.       ,  0.       ,  0.       ,  0.       ]]
```
And there is a corresponding table of reference coordination number values for every element. For H and C:
```
H: [+0.9118, +0.0000, -1.0000, -1.0000, -1.0000]
C: [+0.0000, +0.9868, +1.9985, +2.9987, +3.9844]
```
Negative values are padding. You see that the non-negative values correspond to the non-zero rows and colums of the first matrix. The final C6 coefficient is a Gaussian interpolation of all of the valid reference C6 values.

This calculation corresponds to the set of equations (16) in the paper. I should probably write this into the wiki at some point.

tests/test_vdw.py
NikoOinonen: I mostly wrote this just to test that I get the same result back with both backends, and to have something automated running on Github so that if something changes we can detect it. These test are perhaps not the clearest pieces of example code, so if we want to have coding examples, probably it's better if they would be something separate?

ppafm/cpp/Forces.h
NikoOinonen: It shouldn't be required here. Notice that `r2` (or the higher powers) are not used directly as the divisor here, but rather the damping constants `R0_6` and `R0_8` are added to it before the division. And the damping constants should never be zero.

ProkopHapala: yes, you are right, I did not look carefylly

PR #168, opened 2023-05-30
Set fixed locale during array read.

### LINKED ISSUE #167: Locale problem resurfaced
In a previous PR, we removed a line of code that was changing the locale inside a C++ function because it was causing trouble in Python: #141, #113. Now I ran into an xsf file that is not being read properly by that function, so the electrostatics calculation fails. So there needs to be a more fundamental solution to this problem so that everything works as intended.

NikoOinonen:
Fixes #167 

The locale change was previously removed because it was causing trouble with file reading in Python. However, the removal caused the array read in C++ to fail when using the GUI, because pyqt seems to set a different environment locale, which on some systems uses , (comma) as a decimal separator. This sets a fixed locale temporarily during the array read and then sets it back so that everything should work as intended.

### DIFF
--- a/ppafm/cpp/GridUtils.cpp
+++ b/ppafm/cpp/GridUtils.cpp
@@ -31,6 +31,9 @@ extern "C" {
 
     DLLEXPORT int ReadNumsUpTo_C (char *fname, double *numbers, int * dims, int noline) {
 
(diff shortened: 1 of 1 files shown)


ProkopHapala (approved):
OK, at least we know now why it was there :-)

PR #170, opened 2023-06-12
Release automatically on PyPi

### LINKED ISSUE #107: Use GitHub actions to do a release to PyPi any time a new tag is pushed to the repo.

yakutovicha:
fixes #107

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -36,3 +36,38 @@ jobs:
               with:
                   name: release
                   path: dist/
+
+
+    publish:
+
+        name: Build and publish on PyPI
+        if: startsWith(github.ref, 'refs/tags')
+
+        needs: [build]
+        runs-on: ubuntu-latest
+
+        environment:
+            name: PyPI
+            url: https://pypi.org/project/ppafm/
+
+        steps:
+
+            - uses: actions/download-artifact@v2
+              name: Download distribution artifact
+              with:
+                  name: release
+                  path: dist/
+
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,68 @@
...
+        tags:
+        # Tags that start with the "v" prefix are published on PyPI.
+            - v*
+
+jobs:
+    build:
+        runs-on: ubuntu-latest
+
+        steps:
+
```
NikoOinonen, line 15: Would we not also need a step that installs `g++` and `make`? Or does the `ubuntu-latest` come with these pre-installed?

yakutovicha: I think that `ubuntu-latest` has them. At least when I was trying this out, everything went without a problem.

In the worst case, we fix it in a follow-up PR.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,68 @@
+---
+name: Publish on PyPI
+
+on:
+    push:
+        tags:
+        # Tags that start with the "v" prefix are published on PyPI.
+            - v*
```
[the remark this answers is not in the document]
yakutovicha: Tagging is taken care of by bumpver, #171.

The `bumpver` tool makes a new commit with a version number update and tags it with the same version. Then `bumpver` pushes the commit and the tag to GitHub. Once GitHub receives the tag, it runs the `publish.yml` action to make a new release.

NikoOinonen (approved):

PR #172, opened 2023-06-13
Improve the `README.md` file

### LINKED ISSUE #86: Errors in the installaton description in README.md
As I was compiling the code anew after a fresh clone according to the instructions in README.md, I noticed the following issues with the description there:
1. The compiler complains about missing SDL2/SDL.h unless I have sdl2 available and have set the path to its header files. So I think we should quote SDL2 among the Requirements in the appropriate section of README.md
2. The `cd ProbeParticleModel/Graphene` command before `make` should in fact be `cd ProbeParticleModel/gcc`
I am filing this issue mainly because this relatively innocuous correction of README offers me an opportunity to learn the whole procedure of starting an issue, modifying a file, making a pull request etc. If anyone has time for that, you may check README.md in case you spot some more obvious problems with the instructions for installation and example running, so that we could correct them all at once. Otherwise, let me know if I can proceed with making the Pull request. Anything else that needs to be done? Should I reference this issue in the PPM Roadmap?

yakutovicha:
fixes #86

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,71 +3,38 @@
 
 Simple and efficient **simulation software for high-resolution atomic force microscopy** (**HR-AFM**) and other scanning probe microscopy (SPM) techniques with sub-molecular resolution (STM, IETS, TERS). It simulates deflection of the particle attached to the tip (typically CO molecule, but also e.g. Xe, Cl-, H2O and others).
 
-### Further information
-- Publications: https://github.com/Probe-Particle/ProbeParticleModel#notable-publications-using-probe-particle-model
-- Wiki: https://github.com/Probe-Particle/ProbeParticleModel/wiki
-- API documentation: https://ppafm.readthedocs.io/en/latest/
+## Installation
 
-## Flavors of PPM
+To install the latest version of PPM, run:
 
-Since 2014 PPM developed into the toolbox of various methodologies adjusted for a particular use case.
+```bash
+$ pip install ppafm
+```
 
-1. **CPU version:** - Original implementation using Python & C/C++. It can simulate a typical AFM experiment (3D stack of AFM images) in ~1 minute. It is the base version for the development of new features and methodology. All available simulation models are implemented in this version, including:
-   1. **Point charge electrostatics + Lennard-Jones:** Original fully classical implementation allows the user to set up calculation without any ab-initio input simply by specifying atomic positions, types and charges.
-   2. **Hartree-potential electrostatics + Lennard-Jones:** Electrostatics is considerably improved by using Hartree potential from DFT calculation (e.g. LOCPOT from VASP) and using the Quadrupole model for CO-tip. We found this crucial to properly simulate polar molecules (e.g. H2O clusters, carboxylic acids, PTCDA) which exhibit strong electrostatic distortions of AFM images. Thanks to implementation using fast Fourier transform (FFT) this improvement does not increase the computational time (still ~1 minute), as long as the input electrostatic field is accessible.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,91 +3,83 @@
...
-##### First run: Graphene with point-charges
- 1. clone the repository: `clone https://github.com/Probe-Particle/ProbeParticleModel.git`
- 2. compile the C/C++ modules
-    * `cd ProbeParticleModel/Graphene`
-    * `make`
- 3. Navigate to examples directory `cd ProbeParticleModel/examples/Graphene`this example uses simple (Point-charges + Lennard-Jones)
- 4. run the example `./run.sh`
- 5. output directory `/examples/Graphene/Q-0.05K0.50/Amp2.0` should contain simulated images with tip charge -0.05e, stiffness 0.5N/m and ossicaltion amplitude 2.0A.
+```bash
+$ sudo apt install git python3-pip python3-pyqt5
```
NikoOinonen, line 31: I think we can change this to just
```suggestion
$ sudo apt install python3-pyqt5
```
since you don't need git when installing from PyPI and you should have pip already if you managed to get here.

One thing to maybe note here is that this requires rights to install system packages, which you may not have. There is also the option of using Anaconda, which seems to manage qt5 support on it's own, and does not require administrative access. But it's an additional hurdle, so I would maybe make a wiki page with additional installation information like this.

README.md
```diff
@@ -3,91 +3,83 @@
...
+* CPU: `sudo apt install pocl-opencl-icd`
 
-*NOTE:* Notice that the script `run.sh` downloads and unpack LOCPOT file:
-`wget --no-check-certificate "https://www.dropbox.com/s/18eg89l89npll8x/LOCPOT.xsf.zip"`
-`unzip LOCPOT.xsf.zip`
-this is a large 3D volumetric file which contains Hartree electrostatic potential (in this example computed by VASP) which have to be provided.
+### Use ppafm Docker container
 
-##### Example 3: Pyridine with Density-overlap
+We propose to use [Docker](https://docs.docker.com/get-docker/) to make the code platform-independent.
```
NikoOinonen, line 48: But we also plan to have wheels for all platforms, which should become the more straightforward installation option for every platform. Docker is useful to have as an option, but in order to keep the README minimal, maybe this should also be moved to the wiki page with additional installation options.

README.md
```diff
@@ -3,91 +3,83 @@
...
 ```
 
-###### Usage:
+You can study the script to see how to run the simulation.
+Also, have a look at the `params.ini` file to see how to set up the simulation parameters.
+
+
+Once the simulation is finished, a number of files and folders will be created.
+
+### GUI usage
```
NikoOinonen, line 81: I would move this whole section to the wiki as well. The GUI needs more documentation anyway.

NikoOinonen (commented):
I think some of these sections could be moved to the wiki to keep the README simpler. I could make those page in the wiki and make a separate PR to delete the corresponding sections in the README, if that's okay?

yakutovicha:
> I think some of these sections could be moved to the wiki to keep the README simpler. I could make those page in the wiki and make a separate PR to delete the corresponding sections in the README, if that's okay?


Sure. In that case, nothing is left to do for the current PR. After this is merged, you can go ahead with the follow-ups.

README.md
yakutovicha: Since you proposed to make follow-up PRs, will you take care of that too?

NikoOinonen: Sure, I'll do it.

Issue #106, opened 2023-03-06
Build wheels for different operating systems.


yakutovicha:
Once the release process [is automated](https://github.com/Probe-Particle/ppafm/issues/105), we need to build wheels for different operating systems. This will allow users to not worry about the need for compilation - the CPP code will come pre-compiled already.

NikoOinonen:
There already at least 2 people that have had trouble compiling on Windows (#110), and some difficulties on Linux as well (#135). 

@yakutovicha Could we get some higher priority on this issue? You were setting up things with PyPI before, so I'm assuming you know how to do this. I could also look into it if I just know where to start.

NikoOinonen:
I also ran into this project that promises to automate the builds for all platforms: https://github.com/pypa/cibuildwheel

NikoOinonen:
@yakutovicha  

I read through the PEPs for the `manylinux` tags for the linux wheels:
- `manylinux1`: https://peps.python.org/pep-0513/
- `manylinux2010`: https://peps.python.org/pep-0571/
- `manylinux2014`: https://peps.python.org/pep-0599/
- `manylinux_X_Y`: https://peps.python.org/pep-0600/

It seems that the main rationale for the `manylinux` tags is that different linux distributions have different userspace libraries which limits compatibility for wheels built on different distros. The `manylinux` tags defines a common standard based on the libraries available on different versions of CentOS, with Docker images available.

The standard builds (`python -m build`) don't add these tags, and rather they have to be added manually afterwards. There is a tool called `auditwheel` (https://github.com/pypa/auditwheel) that can be used on a built wheel to check which version of the `manylinux` tag it complies with. In general older versions are desirable, since they are compatible with the most systems.

For the CI build that failed to be uploaded, the version tag that `auditwheel` gives out is `manylinux_2_31_x86_64`, meaning that it's compatible with `glibc` version 2.31 (and newer), which was released in 2020 (https://sourceware.org/glibc/wiki/Glibc%20Timeline). I would target at least as far back as the `manylinux2014` (=`manylinux_2_17`) based on CentOS 7, which is what is running on our local compute cluster.

I think I will try out the [cibuildwheel](https://github.com/pypa/cibuildwheel) that seems to offer CI builds with the different Docker images easily.

NikoOinonen:
[...]

PR #173, opened 2023-06-13
Temporary disable wheel build.

yakutovicha:
Otherwise, the wheel that is created cannot be uploaded on PyPI.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,8 +23,8 @@ jobs:
             - name: Install build
               run: python -m pip install --upgrade build
 
-            - name: Build a binary wheel and a source tarball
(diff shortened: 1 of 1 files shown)


NikoOinonen (approved):
Ok, so the source distribution upload works, but wheel does not. I guess we can resolve it with #106. Do we have any idea why it does that?

yakutovicha:
> Ok, so the source distribution upload works, but wheel does not.

ineed.

> I guess we can resolve it with https://github.com/Probe-Particle/ppafm/issues/106. Do we have any idea why it does that?

It looks to me that the wheel name is something very specific, [see here](https://realpython.com/python-wheels/#what-is-a-python-wheel). The one produced by `python -m build ...` does not match the specification. But I need to look closer at the issue.

PR #174, opened 2023-06-14
Replaced origin='image' with origin='lower' in PPPlot.py

LauriKurki:
Replaced the incorrect keyword argument 'image' with 'lower' in PPPlot.py

I remember these were mostly fixed a while ago but seems that some were left behind.


ondrejkrejci (approved):
I believe that you have tested the thing.

PR #178, opened 2023-06-28
CI wheels build

NikoOinonen:
Adds a job to the PyPI CI workflow that builds wheels for all platforms (64-bit).

The job generates a separate wheel for every Python version `>=3.7`. I don't think there is actually any difference in the wheels between the different Python versions, but I didn't find a way to just generate one. I tested that the wheel build works, but I didn't test how it interacts with the upload to PyPI.

The minimum Python version had to be bumped up to `3.7` because there does not exist a new-enough version of `setuptools` for `3.6`.

Removed the lookup for MSVC variables on Windows because there is a Github action that does it better.


NikoOinonen:
Related to #106, but let's not close it until there is a release on PyPI.

.github/workflows/publish.yml
```diff
@@ -8,7 +8,31 @@ on:
...
+    build_wheels:
+
+        name: Build wheels on ${{ matrix.os }}
+        runs-on: ${{ matrix.os }}
+        strategy:
+            matrix:
+                os: [ubuntu-22.04, windows-2022, macos-11]
+
+        steps:
+            - uses: actions/checkout@v3
```
yakutovicha, line 21: The `build_sdist` contains this step too, please update the `checkout` version to v3 there as well.

yakutovicha (changes requested):
Only a minor change request. The rest looks good.

After this is merged, I suggest we try to make a patch release `0.2.1` and see if things work.

.github/workflows/publish.yml
NikoOinonen: Done.

yakutovicha (approved):
Very cool 😎 !

Issue #148, opened 2023-04-17
Inconsistent parameter loading from script arguments


NikoOinonen:
I noticed recently that the z tip coordinate that is reported on the plotted images was for some reason offset by 0.5Å on the graphene example compared to what I was expecting it to be based on what I did in the OpenCL implementation. I looked into this more closely, and I think the issue is with the parameter loading from the plot script arguments.

The z distance is calculated by adding half of the amplitude to the scan z coordinate at: https://github.com/Probe-Particle/ppafm/blob/8763be1e7e851cb47a1294d49eb04fcd1c14b34b/ppafm/cli/plot_results.py#L239-L242 Notice that it uses the value from the params dict `PPU.params['Amplitude']`. These values are parsed from the command line arguments in the `apply_options` function, but it only accepts keywords that are already in the params dict: https://github.com/Probe-Particle/ppafm/blob/8763be1e7e851cb47a1294d49eb04fcd1c14b34b/ppafm/common.py#L228-L230

However, the command line argument for the amplitude is just `a`: https://github.com/Probe-Particle/ppafm/blob/8763be1e7e851cb47a1294d49eb04fcd1c14b34b/ppafm/cli/plot_results.py#L29 which does not match the value existing in the params dict: https://github.com/Probe-Particle/ppafm/blob/8763be1e7e851cb47a1294d49eb04fcd1c14b34b/ppafm/common.py#L48 so it never gets updated and it uses the default value which is different from the value supplied from the command line.

There are some other command line arguments that have the same issue. It would probably be beneficial to make some kind of dictionary of common command line arguments to use in the scripts to make sure that the names always match.

@ProkopHapala @ondrejkrejci @yakutovicha

ondrejkrejci:
Hi, yeah this is a long way problem, how to define height when the tip is effectivery osscilating. It is the closest poin (I have used that in some paper, but always mentioned explicitely)? Is it the center of the oscillation? To be honest, I am not sure, if there is a standard in the field. Basically we should discuss, choose one of these and use them all around - in the papers, all parts and types of the code.

NikoOinonen:
I'm fine with the current convention where it's the half-point of the oscillation. If that should be changed, then I think that's another topic. The issue that I wanted to address here is the problem with the parsing of the command line arguments into the parameter dict.

yakutovicha:
@mondracek wanted to have a closer look at this issue and share some ideas.

ondrejkrejci:
[...]

PR #179, opened 2023-07-03
Fixes #39

### LINKED ISSUE #39: missing automatic loading of FHI-aims geometry.in and VASP POSCAR files
`pyProbeParticle/basUtils.py` - `loadGeometry(fname=None,params=None)`
https://github.com/Probe-Particle/ProbeParticleModel/blob/a97b04adcfea4228384c4bd946001599c34b4a83/pyProbeParticle/basUtils.py#L440 
no flag for something like **is_in** - meaning is geometry.in file from FHI-aims. It could have names ending on _.in_ or _.in.next_step_.
there is a procedure for reading the geometry.in files here: 
`loadGeometryIN`
https://github.com/Probe-Particle/ProbeParticleModel/blob/a97b04adcfea4228384c4bd946001599c34b4a83/pyProbeParticle/basUtils.py#L130
If file is named _POSCAR_ or _CONTCAR_ we can use this procedure:
`loadPOSCAR`
https://github.com/Probe-Particle/ProbeParticleModel/blob/a97b04adcfea4228384c4bd946001599c34b4a83/pyProbeParticle/basUtils.py#L149
We just need to be sure, that the _atoms_ format (`xyzs` in case of these 2 procedures) is the same as for the other procedures used in loadGeometry.

mondracek:
Fixes #39 

This should address issue #39 (automatic loading of FHI-AIMS geometry.in and VAS POSCAR files). Now, you can do e.g.
`ppafm-generate-ljff -i geometry.in`
`ppafm-generate-ljff -i poscar`
or
`ppafm-generate-ljff -i vasp_file_with_a_funny_name --format poscar`
and the `ppafm` code recognizes the input geometry format. Valid options for the `format` argument are `xyz`, `in`, `poscar`, `xsf`, and `cube`.

We should probably add some automatic tests for this functionality but I have not come up with anything yet. Obviously, the natural way is to create some example files in the `geometry.in`, `atoms.xyz` and `POSCAR` formats and see if they load correctly. But to include this into automatic testing, we would essentially need to create new example folders and distribute them with the code, right? Do we want to do it just to have some tests available?


ppafm/io.py
```diff
@@ -374,15 +389,19 @@ def loadGeometry(fname=None,params=None):
...
+        atoms, nDim, lvec = loadXSFGeom(fname)
+    elif(format=="npy"):
+        atoms, nDim, lvec = loadNPYGeom(fname) # under development
+    elif(format=="poscar"):
+        atoms, nDim, lvec = loadPOSCAR(fname)
+    elif(format=="in"):
+        atoms, nDim, lvec = loadGeometryIN(fname)
+    #TODO: introduce a function which reads the geometry from the .npy file
     else:
         sys.exit("ERROR!!! Unknown format of geometry system. Supported formats are: .xyz, .cube, .xsf \n\n")
```
NikoOinonen, line 406: The message is wrong (we support more formats), and it would be better if this was a `raise ValueError()` instead of `sys.exit()`, so that if you use this function in some place that can catch the error, it won't just suddenly exit the whole program.

ppafm/cli/generateDFTD3.py
```diff
@@ -22,6 +22,7 @@ def main():
 
     parser.add_argument('-i', '--input',       action='store',
         help='Input file. Mandatory. Supported formats are: .xyz, .cube, .xsf.')
+    parser.add_argument("--format", action="store", type="string", help="Format of the input geometry file (overrides format concluded from the file name extension)", default=None)
     parser.add_argument('-f', '--data_format', action='store', default='xsf',
```
NikoOinonen, line 26: I think it would better to reconsider the names of the arguments. The `format` and `data_format` may be confusing for a new user. Maybe `input_format` and `output_format` instead.

But I think the CLI arguments need an overhaul anyway, so this could be done separately. Also related to #148.

NikoOinonen (changes requested):
A couple of things to fix, otherwise looks good.

NikoOinonen:
> But to include this into automatic testing, we would essentially need to create new example folders and distribute them with the code, right? Do we want to do it just to have some tests available?

We have a dedicated folder called `tests` in the root of the repo. This folder is not distributed with the rest of the code with pip, so we are free to add anything we want there and it won't interfere with the users. I don't think it would be an issue to add a couple of example input files there (maybe in a subfolder), since these file formats are very small in size.

ppafm/cli/generateDFTD3.py
NikoOinonen: I'm working on overhauling the CLI arguments as a part of resolving #148, so let's not touch this here.

mondracek:
The typo and the error messages pointed out by @NikoOinonen were addressed in the last commit(s). Niko, I will now wait till you resolve #148, so that we can decide what the name of the argument for input format specification shall actually be ("input_format" perhaps?).

mondracek:
The name of the argument that specifies the input data format is now (after merging work done in PR #185)
`--input_format`, short version `-F`.
E.g.
```ppafm-generate-ljff -i geometry.out -F in```
or
```ppafm-generate-elff -i LOCPOT.xsf~ --input_format xsf```
etc.
The argument is not case sensitive (`XSF` and `xsf` would be the same).

While integrating this into `generateElFF`, I stumbled over how the format of some other input files, those passed through the `--tip_dens`, `--KPFM_tip`, and `--KPFM_sample` arguments, should be treated. The latter two files may be needed for KPFM, the first would be a tip charge density  from _first principles_, not necessarily for KPFM only. As for now, the format of these files is determined based on the file name extension, which I guess should stay that way, no need to add an analogy of the `--input_format` for each of them. I have just added the possibility of `tip_dens` to be in the `CUBE` format and, furthermore, I have added checks that all these files are either `*.cube` or `*.xsf` and that the `KPFM_tip` and , `KPFM_sample` are both in the same format as the main input file (specified by `-i`) , otherwise an error is issued. But as the KPFM is not working at the moment, these additions remain untested.

**Comment by mondracek** [comment]
**Trouble with `flake8` when committing**
I have made the last commit with `git commit -n`, that is, with the automatic tests turned off. Unless I do this, I get *lots* of complaints from `flake8` about various formatting issues and the test fails. I actually find most of the formatting suggestions by `flake8` reasonable, but
- there are really many of them and
- most of them are not even about the code I have written myself as a part of this specific branch/PR,

so I do not think this here is the right place to implement them. I can start a new issue and split off a dedicated branch, if needed. Anyway, I wonder, am I the only one who hit into this issue? Is everybody making *commits* with the `-no-verify` option?

**Comment by NikoOinonen** [comment]
>  Is everybody making commits with the -no-verify option?

Probably yes, because there are _a lot of_ problems in the current code that flake8 does not approve of. Actually most of these are code style things, but there is no consensus on whether we want to enforce a style or not. For now you just have to use the `--no-verify` option. (Actually, we probably should disable the flake8 test until there is some agreement on it...)

**Comment by NikoOinonen** [comment]
@mondracek Is this ready for review? If it is, you can re-request the review from the right side using the circling arrows button next to my user name.

File: ppafm/cli/generateElFF.py [hunk 3]
```diff
@@ -77,14 +88,18 @@ def main():
...
     if args.tip_dens is not None:
         ###  NO NEED TO RENORMALIZE : fieldFFT already works with density
-        print(">>> loading tip density from ",args.tip_dens,"...")
-        rho_tip, lvec_tip, nDim_tip, head_tip = io.loadXSF( args.tip_dens )
-
+        print(">>> Loading tip density from ",args.tip_dens,"...")
+        if(args.tip_dens.lowercase().endswith("xsf") ):
+            rho_tip, lvec_tip, nDim_tip, head_tip = io.loadXSF( args.tip_dens )
+        elif(args.tip_dens.lowercase().endswith("cube") ):
+            rho_tip, lvec_tip, nDim_tip, head_tip = io.loadCUBE( args.tip_dens )
```
NikoOinonen, line 95: Actually, this doesn't work, because the `loadCUBE` function automatically multiplies the array by the conversion factor between Hartree and eV, but here we have an electron density instead of a Hartree potential. I would remove this for now. If we want to support .cube also for electron densities, we need to fix the problem with the factor, which is probably another issue.

NikoOinonen (changes requested):

NikoOinonen (approved):
Ok, looks good now.

PR #180, opened 2023-07-05
Fixed API docs for AFMulator.

NikoOinonen:
The API documentation page for `AFMulator` was missing items.


ProkopHapala (approved):
I don't undrstand it, but I think such small and simple changes should go forward without waiting for review

PR #181, opened 2023-07-21
Update the docker container.

yakutovicha:
Slight update of the Dockerfile to be in sync with the latest developments of ppafm.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
                   path: ./*.png
 
     test-docker:
-        name: Build docker image on ${{ matrix.os }}
+        name: Test docker image on ${{ matrix.os }}
         runs-on: ${{ matrix.os }}
         strategy:
             matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
             - name: Execute the container.
               run: |
                   cd examples/PTCDA_single/
-                  docker run --rm -u $(id -u):$(id -g) -v ${PWD}:/exec ppafm:latest ppafm-generate-ljff -i PTCDA.xyz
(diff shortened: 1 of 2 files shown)


NikoOinonen:
I was just testing this on my computer, and got the same `PermissionError: [Errno 13] Permission denied: 'FFLJ_x.xsf` as in the CI test. For me the problem was that the example folder did not have write permission for other users. When I set the `777` permissions, then it worked.

yakutovicha:
> I was just testing this on my computer, and got the same `PermissionError: [Errno 13] Permission denied: 'FFLJ_x.xsf` as in the CI test. For me the problem was that the example folder did not have write permission for other users. When I set the `777` permissions, then it worked.

Yes, I just tried that approach too and added to the test.

yakutovicha:
By the way, @NikoOinonen, do you have a win machine? That would be great if someone could test it on win too.

NikoOinonen:
Sure, I can try on Windows as well.

NikoOinonen:
Also getting that setuptools error now. I have seen this before, when pip for some reason priorities the Ubuntu system setuptools installation which is some ancient version. It can be fixed in a kind of brute force way by removing the default setuptools installation and replacing it like here: https://askubuntu.com/questions/683601/how-to-upgrade-python-setuptools-12-2-on-ubuntu-15-04/683730#683730

yakutovicha:
@NikoOinonen I finally managed to simplify the Docker file and make things work. The PR is ready for your review and testing.

yakutovicha:
Maybe we should test this docker image for different systems as well, what do you think @NikoOinonen?

yakutovicha:
Well, I tried and failed. I don't have too much interest and time to work on this further. Let's review/merge it as it is. We can always add more stuff later.

NikoOinonen:
It seems to work in both Linux and Windows. On Windows the only difference is that in cmd `PWD` does not work, so you run docker with
```bash
docker run --rm -v %CD%:/exec ppafm:latest ...
```
I added a note of that to the [wiki](https://github.com/Probe-Particle/ppafm/wiki/Additional-installation-instructions).

NikoOinonen (approved):

Issue #146, opened 2023-04-14
Proper documentation of the new executable scripts


NikoOinonen:
Some of the scripts that we previously had in the root folder are now accessible under a different name in the pip installed version (e.g. `generateLJFF.py` becomes `ppafm-generate-ljff`). The documentation should be updated to reflect this change. At least the quick start guide on the wiki (https://github.com/Probe-Particle/ppafm/wiki#running-calculations) and the dedicated page on the scripts (https://github.com/Probe-Particle/ppafm/wiki/Root-Scripts) need to be updated.

ondrejkrejci:
[Root-Scripts](https://github.com/Probe-Particle/ppafm/wiki/Root-Scripts) updated. One PR will be connected with 1 missing script. 
@NikoOinonen  - please have a look on both Wiki's and tell me if you dislike something.
@yakutovicha - please have a look on description of your [Fitting Scripts](https://github.com/Probe-Particle/ppafm/wiki/Root-Scripts#fitting). Thank you!

NikoOinonen:
Good job with the root scripts wiki page. I fixed a couple of typos and things, but otherwise very good. The only thing that stood out  to me was the phrase
> [Lennard-Jones](https://en.wikipedia.org/wiki/Lennard-Jones_potential) force field, which captures the interaction of a weakly-interacting particle

I would argue that the interaction is not that weak in the repulsive range. Or does this maybe refer to the fact that it doesn't model a chemical reaction?

yakutovicha:
> @yakutovicha - please have a look on description of your [Fitting Scripts](https://github.com/Probe-Particle/ppafm/wiki/Root-Scripts#fitting).

Looks good to me

yakutovicha:
Two pages got my attention: [MAC users](https://github.com/Probe-Particle/ppafm/wiki/MAC-users) and [Windows Users](https://github.com/Probe-Particle/ppafm/wiki/Windows-Users).

Provided that their content is outdated and now we have [notes on different platforms](https://github.com/Probe-Particle/ppafm/wiki/Additional-installation-instructions#notes-on-different-platforms) I suggest that "MAC-users" and "WIndows Users" can be removed.

Do you agree, @ondrejkrejci and @NikoOinonen?

NikoOinonen:
[...]

PR #182, opened 2023-07-25
Adding fit size script to Utilities

ondrejkrejci:
Connected with #146 .
I have added a script into utilities, which was there some time ago, and tested it with the new code. No automatic test added as it is external script. The script is added as it was used in my recent publication.
I have found some other mistakes in different scripts, so I have adjusted them.
The scripts are now described in https://github.com/Probe-Particle/ppafm/wiki/Root-Scripts .


NikoOinonen (approved):
I didn't test the new script, but otherwise looks good. And good job on the wiki page.

PR #185, opened 2023-07-29
Overhaul of CLI arguments

### LINKED ISSUE #148: Inconsistent parameter loading from script arguments

NikoOinonen:
fixes #148 

Adds to `ppafm.common` a new class `CLIParser`, which holds a list of common commandline arguments that can be added to the scripts. E.g.
```python
parser = CLIParser()
parser.add_arguments(['output_format', 'Amplitude', 'charge', 'noPBC'])
```
The arguments are divided into two categories: those which directly correspond to one of the parameters in the `ppafm.common.params` dictionary, and others. The parameter arguments are checked so that their names match with the names in the parameter dictionary to avoid problems.

All the main CLI scripts are changed to use the new parser. But not all arguments are transferred to the pre-defined list; only the ones which are used in more than one place.

A couple of the arguments are renamed:
- `data_format` -> `output_format` to contrast with the added new argument `input_format`. @mondracek 
- `Bpower`-> `Bpauli` in order to be consistent with the parameter name in the params dictionary.
- `densityMayBeNegative` -> `no_negative_check` in `conv_rho.py` for a more descriptive name.


ppafm/common.py
```diff
@@ -57,11 +58,174 @@
     'tip_base':  np.array( ['None', 0.00 ]),
     'Rtip'         :  30.0,
     'permit'       :  0.00552634959,
-    'Vrange':   0.0,
     'vdWDampKind' : 2,
     '#' : None
 }
 
+class CLIParser(ArgumentParser):
```
ProkopHapala, line 65: Aha it is extension of ArgumentParser. Is it some common practice to make you own derived class of ArgumentParser?

ProkopHapala (approved):
I perhpas did not encounter the problem which this should solve, so I don;t fully see it it is worth it. But I believe you have good reason to do this.

I'm thinking about that "ArgumentParser" is some standard of-the-shelf python module which everybody knows and use. Now we modify it to ower own custom made solution, which may be a bit more confusing to people. But perhaps it is worth it.

ppafm/common.py
NikoOinonen: It's maybe not common practice, but I think a good compromise. The main concern for me was that the arguments were not consistent with the names in the `params` dictionary. I wrote about the example of the amplitude being read wrong in one of the scripts in the associated issue #148. If we add the arguments through this interface, it's less error prone, there is less repeated code, and the help messages are more consistent.

NikoOinonen:
It's approved, but I'll wait for @mondracek to give his opinion, since the CLI arguments concern his other PR #179.

mondracek:
Seems good to me. Just one thing: When running the example in ``ppafm/examples/PTCDA_Hartree_dz2``, the line
```ppafm-plot-results -k 0.5 -q -0.10 -a 2.0 2 --df -f npy```
in ``run.sh`` results in the following error:
```ppafm-plot-results: error: unrecognized arguments: 2```
As I understand it, this is not really an error introduced by the new code. Rather, the value ``2`` for the amplitude is given twice by mistake in the above line. The old way of handling the arguments probably failed to notice this mistake and tolerated the example script as it is. Or perhaps it did not and I just did not notice that the example was not working before. Anyway, I suggest @NikoOinonen that you correct this line in ``ppafm/examples/PTCDA_Hartree_dz2/run.sh`` and commit it to this branch. Then, let's proceed with merging to ``main``. Once this PR #185 is merged, I will go on with my issue #39.

mondracek:
Okay, I think as this `` examples/PTCDA_Hartree_dz2/run.sh`` issue is literally one character in one file, I can as well repair it myself and push the commit. So I am going to do it now.

PR #187, opened 2023-08-04
Temporarily disable flake8 pre-commit hook.

NikoOinonen:
I'm suggesting that we temporarily disable the `flake8` pre-commit hook, since there has been no consensus on whether `black` should be enabled or not (#96). At the moment the pre-commit never succeeds because of all the code style problems, so it's mostly just an annoyance since in practice you have to use the `--no-verify` option every time or manually disable the check locally.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,20 +24,20 @@ repos:
       hooks:
           - id: pyupgrade
             args: [--py36-plus]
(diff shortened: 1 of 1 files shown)


mondracek (approved):
Let's do it. But we should reactivate it later when we are ready to review all the white-space formatting issues. Comparing and merging too branches that do not differ much but differ a lot in the white spaces is a hell.

PR #188, opened 2023-08-04
Merge relaxed_scan_PVE.py to relaxed_scan.py.

### LINKED ISSUE #186: Merge `relaxed_scan_PVE.py` into `relaxed_scan.py`?
While doing #185, I noted that the `relaxed_scan_PVE.py` script used in the density overlap model is almost identical to the standard `relaxed_scan.py` except for a couple of additions.

@ProkopHapala Do you think we should merge these two scripts into one? It seems like this could be done quite easily just by adding some argument like `--noLJ` to `relaxed_scan.py`.

NikoOinonen:
Fixes #186 

Transfers all functionality from `relaxed_scan_PVE.py` to `relaxed_scan.py` and deletes `relaxed_scan_PVE.py`


mondracek (commented):
I got temporarily lost in the differences between the versions and spent a whole day on it, sorry. I am nearly finished reviewing the stuff now, but I may need to postpone the final word to Monday. I will probably suggest a few changes, but nothing dramatic.

ProkopHapala (approved):
great

mondracek:
My latest commits contain just some not very important formatting issues. My idea was to revise the differences between the old `relaxed_scan.py` and `relaxed_scan_PVE.py`, including even the small differences in formatting (of which there were surprisingly many) and adopt the variant which I liked better in the new `relaxed_scan.py`. Probably too much work of too little utility, but it's done so it's done. However, I wanted to propose two more thing. Unfortunately, @NikoOinonen already merged the `merge-relaxed-scan` branch as I am typing this, but I will mention here what I wanted to do anyway, in the next comment.

mondracek:
First, I agree that the newly introduced `--noLJ` is the right way to distinguish between the regime which uses the Lennard-Jones potential and the one which reads the Pauli and the separate vdW-only force fields. However, for the sake of user-friendliness, I would prefer to also have the feature that when the user forgets to specify the `--noLJ` option but his/her working directory lacks the `FFLJ_*` files while it does contain the `FFPpauli_*` and `FFvdW_*` files, the program still defaults to using the separate Pauli and van der Waals force fields.

Second, the present version of the `ppafm/examples/pyridineDensOverlap/run.sh` script mixes two ways of calling the `ppafm` scripts: It calls ``python ${PPAFM_DIR}/generateElFF.py``, using the explicitly pre-defined path variable ``PPAFM_DIR='../../ppafm/cli'``, but then calls the `pip`-installed scripts ``ppafm-relaxed-scan`` and ``ppafm-plot-results``. I think we should keep ``python ${PPAFM_DIR}/relaxed_scan.py`` and ``python ${PPAFM_DIR}/plot_results`` here for consistency. This would serve as an example for the user that such way of running the script is also possible. As this older way is currently required for running ``python ${PPAFM_DIR}/conv_rho.py`` and ``python ${PPAFM_DIR}/generateDFTD3.py`` anyway, this is an opportunity to have an example that does not need the `pip` installation. Alternatively, we can call the `pip`-installed `ppafm-*` scripts for everything, but unless I missed it (because of my outdated code installation), we have yet to add the ``ppafm-conv-rho`` and ``ppafm-generate-dftd3`` scripts to our package.

I have a ready `alt-relaxed-scan` branch that implements the default `--noLJ` in case of missing `FFLJ_*` files and modifies the `ppafm/examples/pyridineDensOverlap/run.sh`, so I can push it if you like the ideas.

PR #192, opened 2023-08-08
Add ppafm-conv-rho and ppafm-generate-dftd3. Closes #186

### LINKED ISSUE #186: Merge `relaxed_scan_PVE.py` into `relaxed_scan.py`?

mondracek:
Scripts `ppafm-conv-rho` and `ppafm-generate-dftd3` can be used to call `conv_rho.py` and `generateDFTD3.py`, respectively, now. The script `example/pyridineDensOverlap/run.sh` was updated. Problem of undefined `geometry_format` in `getAtomsWhichTouchPBCcell` resolved as testing the `pyridineDensOverlap` example. Done in the aftermath of and should close #186.


ppafm/cli/conv_rho.py
```diff
@@ -21,63 +22,63 @@ def handleNegativeDensity( rho ):
     rho[rho<0] = 0
     rho *= ( Q/rho.sum() )
 
-# ======== Main
+def main():
```
ProkopHapala, line 25: So the intend here is to be able call this as a function from another script? Or what is the purpose/advantage to encapsulate everything inside main function?

ProkopHapala (approved):
I assume it works. Otherwise I don't see much for dicussion. Thanks.

ppafm/cli/conv_rho.py
mondracek: I do not much about how exactly this works, I just followed the template indicated by @NikoOinonen and used e.g. for `ppafm-generate-elff`, but... if you look inside the script `ppafm-conv-rho`, which gets installed by `pip` based on `pyproject.toml`, this script actually loads `conv_rho.py` as ``from ppafm.cli.conv_rho import main`` and then calls it by ``sys.exit(main())``. I have no idea if it could be done in a different way.

NikoOinonen: There is a feature in setuptools called [_entry point_](https://setuptools.pypa.io/en/latest/userguide/entry_point.html), which allows easily adding executable scripts to the package distribution. The idea is that you define a function somewhere in your package, and then you automatically generate an installable script that simply imports and calls that function.

ppafm/cli/conv_rho.py
```diff
@@ -1,15 +1,16 @@
...
 import ppafm.fieldFFT as fFFT
 from ppafm import io
 
-# ======== Functions
+HELP_MESSAGE = f"""Use this program in the following way:
+ppafm-conv-rho -s <filename> -t <filename> [ -A <value> ] [ -B <value> ]
+Supported file fromats are:
+    * xsf
+"""
+
```
NikoOinonen, line 13: I don't think this is needed. You can get a help message for the script by simply calling it with the `-h` option. E.g. `ppafm-conv-rho -h`.

NikoOinonen:
I checked that it works. Naturally it requires a reinstall for the new scripts.

ppafm/cli/conv_rho.py
mondracek: Alright. However, `generateElFF.py` does have the `HELP_MESSAGE` constant defined like the above. Is there any reason why `generateElFF.py` is special? Or should I remove the `HELP_MESSAGE`  from `generateElFF.py` too?

NikoOinonen: I don't see any reason why `generateElFF.py` would need it either. If you need to print the help message from within the program, you can use the [`parser.print_help()`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.print_help) method. I would remove this from both of the scripts.

NikoOinonen:
@mondracek I guess we can merge this now, if you don't have any other concerns? I think we should also make a new release after this is merged, so that the examples work correctly on the release version.

mondracek:
Good, squashing and merging `alt-relaxed-scan` into `main`. Regarding the new version, should it be `v0.2.2` or something else? Anyway, I will leave it up to @NikoOinonen or others to make the new release version.

NikoOinonen:
New release 0.2.2 is out: https://github.com/Probe-Particle/ppafm/releases/tag/v0.2.2

Issue #133, opened 2023-04-04
Application for KPFM


Aimannor:
Can the model be used for modeling KPFM for charged surfaces?

aureliojgc:
Hi,
The link you had send Niko is not the version of KPFM that I implemented. About the information of my version. I think I did not add it to the wiki yet, my bad. I will prepare something ASAP and add it.

aureliojgc:
about the charged surfaces... As long as you can calculate the charged state with DFT and you assume that the extra charge in the surface will not induce any charge transference between the tip and sample and that the tip will not induce any charge rearrangement on the surface it should be fine...

Aimannor:
Thank you for getting back to me so quickly. I have more related questions to ask.
1) Can the model be used for AM or FM KPFM modeling?
2) The other thing is, if I have 2D surface potential and topography contour can I convert them to charge state contour for unknown material using the model?

aureliojgc:
Hi! I am sorry I missed the last messages. I hope you didn't need the answer much in a hurry.

About the modulation, the AFM code only works on FM and since the KPFM is an extension of it, it will only work on FM. maybe @ProkopHapala can spreed further light on this.

and for the surface potential... we never tried something like it as long as I am aware... but if you find the way to extend your data to an electrostatic potential in a tridimensional cube of xsf it will maybe go through. but will be unknown territoty as long I know.

Sorry again for the late answer,
A.

mondracek:
**The KPFM functionality is not working!**
I am attaching an example in which I have attempted to calculate the LCPD on 2 atomic layers of NaCl on the Au(001). The attached archive file contains the inputs (``params.ini``, ``hatree_potential.cube`` with the electrostatic potential of the sample, and ``hartree_potential_Vz.cube`` with the electrostatic potential of the sample in an external field), the script ``run.sh`` with commands to run and the output file ``stdout+err.txt`` that contains the standard output and standard error output messages.
[NaCl-on-Au001.tar.gz](https://github.com/Probe-Particle/ppafm/files/12231656/NaCl-on-Au001.tar.gz)
What I figure is happening: The command
```ppafm-relaxed-scan --Vrange -0.5 0.5 3```
should write the forces into
```Q-0.20K0.25V-0.50/OutFz.xsf```,
```Q-0.20K0.25V0.00/OutFz.xsf``` and
```Q-0.20K0.25V0.50/OutFz.xsf```,
according to the bias (_V_), where the subsequent command
[...]

PR #193, opened 2023-08-21
Making KPFM work

mondracek:
This PR partially addresses issue #133, but we should at least review the KPFM documentation before we close the issue. The KPFM actually already worked after PR #188. Here, I just made sure the `--WSxM` option is not needed if the WSxM format is not required and I renamed the `LCPD_HzperV.xsf` output file to just `LCPD.xsf`, because the LCPD should actually be expressed in volts, not in Hz/V. I wonder, was the `LCPD_HzperV.xsf` name a vestige of some older setup in which `LCPD_b` was written out insted of `LCPD`? (`LCPD_b` is a map of the linear coefficients _b_ in the Kelvin parabola expansion) @aureliojgc , do you know?
I would like to implement more improvements to the KPFM part of the code but I should probably open new issue(s) about it.

### DIFF
--- a/ppafm/PPPlot.py
+++ b/ppafm/PPPlot.py
@@ -81,18 +81,19 @@ def plotImages(
     atoms=None,
     bonds=None,
     atomSize=default_atom_size,
-    symetric_map=False,
+    symmetric_map=False,
     V0=0.0,
 ):
     for ii, i in enumerate(slices):
         # print(" plotting ", i)
         write_plotting_slice(i)
-        if symetric_map:
+        if symmetric_map:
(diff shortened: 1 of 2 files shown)


aureliojgc (commented):
Thanks for the work Martin. Yes, the old name was probably vestigial from earlier versions where I printed the b parameter. But now the LCPD is ofc the LCPD maxima in V.

ProkopHapala:
@mondracek @aureliojgc 
What is going on here? This is pretty old pull request. Why it id not procede?

aureliojgc (approved):
All good by me

NikoOinonen:
@mondracek Could you resolve conflicts and proceed with the merge. I believe the conflicts are just formatting changes from the recent PRs.

PR #196, opened 2023-08-30
Apply black to the `ml` package.

yakutovicha:


NikoOinonen:
How should we do this? Should we add our own commits to fix the things we have issues with, or mark them here on Github and @yakutovicha applies the fixes?

yakutovicha:
> 

you can just mark the place you don't like and I will disable formatting for them

yakutovicha:
Of course, the alternative is to increase the line width to 120 characters. This we can also do.

yakutovicha:
> Actually, I thought it would be 120. I would put it at least that high.

Done, 120 is now the default. See 2c06483fdc31aef85bfa1ad0976492cfcab11ddb for the effect.



> No, I meant just in the Github review comment you can suggest a code commit.

Sure, that is fine with me too.

ProkopHapala:
> would put it at least that high. Most screens can hold 150+ chars in half a screen these days (of course it becomes hard to read at some point anyway

Automatic breaking of long lines is some feature I would perhaps switch of completely.

If you don't want to switch it of I would increase the limit very much (180 e.g. ?). I agree screen today are large enough. And I also prefer horizontal scrolling rather than having broken lines which perturb the code flow (it visually disturbs a lot when you are trying to see big-picture i.e. overall structure of the code)

NikoOinonen (changes requested):
Just a few things stand out to me. Maybe Prokop has more to say about the aligned equals signs in his Corrector code.

NikoOinonen:
> If you don't want to switch it of I would increase the limit very much (180 e.g. ?). I agree screen today are large enough. And I also prefer horizontal scrolling rather than having broken lines which perturb the code flow (it visually disturbs a lot when you are trying to see big-picture i.e. overall structure of the code)

I guess it's a preference thing. I don't really like horizontal scrolling that much, and very long lines can make the cursor jump a lot when navigating by keyboard. I mostly threw the 150 figure because that's about as much text as fits on screen when I code, usually in half screen or a vertically mounted screen. Mostly I wanted to say that the default 80 is way too low for modern equipment.

yakutovicha:
Can we find a compromise about the line length? I am fine to make it 150 or whatever. But I agree with Nico that horizontal scrolling is not the best.

Maybe, if the line is that long, it probably calls for an actual code rewriting.

@ProkopHapala would 150 characters be acceptable for you?

NikoOinonen:
180 is maybe bit on the high side, but it's fine with me.

yakutovicha:
@ProkopHapala, we are waiting for your decision. We are OK with both: 150 or 180. Please let us know the number so I can continue with the work.

ProkopHapala:
> @ProkopHapala, we are waiting for your decision. We are OK with both: 150 or 180. Please let us know the number so I can continue with the work.

Aha, OK, sorry. So 180

yakutovicha:
> > @ProkopHapala, we are waiting for your decision. We are OK with both: 150 or 180. Please let us know the number so I can continue with the work.
> 
> Aha, OK, sorry. So 180

done ✅

yakutovicha:
@NikoOinonen I applied the requested changes, please re-review.

NikoOinonen (approved):
Looks fine to me now.

PR #197, opened 2023-09-11
Create automatic lattice vectors if none are specified in the input

### LINKED ISSUE #52: Handling missing lattice vectors

Some previous suggestions:

From @ondrejkrejci 
> 
> 
> 
>     * set:
> 
> 
> ```
> gridA 0.0 0.0 0.0
> gridB 0.0 0.0 0.0
> gridC 0.0 0.0 0.0
> ```
> 

From @NikoOinonen 
> 

mondracek:
This can close #52
If the lattice vectors are specified neither in `params.ini` (as `gridA`, `grdB`, `gridC`) nor in the geometry input file (typically `input.xyz` but the FHI-AIMS's `geometry.in` can be, in principle, also given without lattice vectors), the code should propose suitable cell dimensions automatically and work with them thereafter. The logic followed to create the automatic cell is this: it should be large enough to encompass all atoms as well as the whole scanning space defined by `scanMin` and `scanMax`, plus some padding. I have currently put the hard-wired value of `3.0` (angstrom) for the padding in there but we can change it or make it more flexible if anyone prefers so.
The piece of code to implemented this is rather easy. My problem was where to place it to make it work reliably. I eventually decided to make it part of the `loadGeometry` function in `ppafm/io.py`. Implementing it as a separate function which gets called after `loadGeometry` might have been a somewhat cleaner solution regarding the code structure, but we would probably want to call it *every time* after `loadGeometry` anyway and would just have to always remember doing it, so I thought we might as well spare ourselves the hassle.
The main piece of the added code is here https://github.com/Probe-Particle/ppafm/blob/052ad8ed6dd5f686557f0895718d65cf479c7dca/ppafm/io.py#L490-L548
In my recent comment under #52, I have complained about a complicated code structure when it comes to handling the `lvec` array, the `grid` parameters etc. The most troubling feature with respect to implementing this part was the repeated shuffling of information about lattice vectors between the local variable(s) `lvec` and the global parameters `gridA`, `gridB`, `gridC`. I tried really hard to figure out when I should work with the former and when with the latter, but I gave up when I was still not sure after an insanely long time. (I do not think the time was completely wasted, it was probably necessary that I spend it to really understand the structure of the code, and why we should probably make it better.) Anyway, as you can see, I finally made this new snippet of code a bit more complex than strictly necessary for its core function, just to make really really sure that the `grid` parameters and the lattice vectors returned as `lvec` are mutually consistent once the work is done.
In my definition of the default lattice vectors, I treat the periodic and non-periodic cases differently. Essentially, the difference between the minimal and maximal coordinates that should be inside the cell defines the (length of) the lattice vectors in the periodic case. For a non-periodic structure, zero is always taken as the minimal coordinate. This is because the origin of the cell will be zero in any case, which does not really matter for periodic structures but makes a difference for non-periodic structures. In the non-periodic structures, anything with negative coordinates becomes inaccessible. This is actually described on our Wiki page under the caption **Cell definitions**, so it is strictly speaking not a bug, but I would still like to do something to alleviate this restriction eventually. So far, however, I am leaving it as it is.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.10
+FROM ubuntu:22.04
 
 # build-essential is needed for building the C++ code in the PPAFM package
 
--- a/ppafm/HighLevel.py
+++ b/ppafm/HighLevel.py
@@ -18,177 +18,290 @@ Vmax_DEFAULT = 10.0
 
 # ==== PP Relaxation
 
+
 def Gauss(Evib, E0, w):
-    return np.exp( -0.5*((Evib - E0)/w)**2);
+    return np.exp(-0.5 * ((Evib - E0) / w) ** 2)
 
-def symGauss( Evib, E0, w):
-    return Gauss(Evib, E0, w) - Gauss(Evib, -E0, w);
 
-def meshgrid3d(xs,ys,zs):
-    Xs,Ys,Zs = np.zeros()
-    Xs,Ys = np.meshgrid(xs,ys)
+def symGauss(Evib, E0, w):
+    return Gauss(Evib, E0, w) - Gauss(Evib, -E0, w)
 
-def trjByDir( n, d=[0.0,0.0,PPU.params['scanStep'][2]], p0=[0,0,PPU.params['scanMin'][2]] ):
-    trj = np.zeros( (n,3) )
-    trj[:,0] = p0[0] + (np.arange( n )[::-1])*d[0]
-    trj[:,1] = p0[1] + (np.arange( n )[::-1])*d[1]
-    trj[:,2] = p0[2] + (np.arange( n )[::-1])*d[2]
+
+def meshgrid3d(xs, ys, zs):
+    Xs, Ys, Zs = np.zeros()
+    Xs, Ys = np.meshgrid(xs, ys)
+
+
+def trjByDir(
+    n, d=[0.0, 0.0, PPU.params["scanStep"][2]], p0=[0, 0, PPU.params["scanMin"][2]]
+):
+    trj = np.zeros((n, 3))
+    trj[:, 0] = p0[0] + (np.arange(n)[::-1]) * d[0]
+    trj[:, 1] = p0[1] + (np.arange(n)[::-1]) * d[1]
+    trj[:, 2] = p0[2] + (np.arange(n)[::-1]) * d[2]
     return trj
 
-def relaxedScan3D( xTips, yTips, zTips, trj=None, bF3d=False ):
-    if(verbose>0): print(">>BEGIN: relaxedScan3D()")
-    if(verbose>0): print(" zTips : ",zTips)
-    ntips = len(zTips);
-    rTips = np.zeros((ntips,3))
-    rs    = np.zeros((ntips,3))
-    fs    = np.zeros((ntips,3))
-    nx = len(zTips); ny = len(yTips ); nz = len(xTips);
-    if( bF3d ):
-        fzs    = np.zeros( ( nx,ny,nz,3) );
+
+def relaxedScan3D(xTips, yTips, zTips, trj=None, bF3d=False):
+    if verbose > 0:
+        print(">>BEGIN: relaxedScan3D()")
+    if verbose > 0:
+        print(" zTips : ", zTips)
+    ntips = len(zTips)
+    rTips = np.zeros((ntips, 3))
+    rs = np.zeros((ntips, 3))
+    fs = np.zeros((ntips, 3))
+    nx = len(zTips)
+    ny = len(yTips)
+    nz = len(xTips)
+    if bF3d:
+        fzs = np.zeros((nx, ny, nz, 3))
     else:
-        fzs    = np.zeros( ( nx,ny,nz   ) );
-    PPpos  = np.zeros( ( nx,ny,nz,3 ) );
+        fzs = np.zeros((nx, ny, nz))
+    PPpos = np.zeros((nx, ny, nz, 3))
     if trj is None:
-        trj=np.zeros((ntips,3))
-        trj[:,2]=zTips[::-1]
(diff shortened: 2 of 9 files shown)


mondracek:
Besides the essence of issue #52, which is handled by the code referenced in my above comment, there are a few other small changes included in this PR, which are either related to the issue or which I simply stumbled over when implementing it and I did not think they are worth separate issues or branches. You can essentially see their list in the long name of commit ebaefb2126cac63cd7edf5ae548416bd0214279a:
- The number of grid points in `gridN` used to be automatically set by calling the `autoGridN` function inside `loadParams` if it had not been specified in `params.ini`. However, the lengths of the lattice vectors should be known to set the automatic grid correctly and only the lattice vectors from `params.ini` would be known at that point. Should the lattice vectors be actually taken from the geometry file (or automatically generated), they were going to change later. The function `autoGridN` called from `loadparams` would not be able to take this into account. I have therefore removed this call to `autoGridN` and I have instead implemented the automatic grid together with the automatic lattice vector creation in `loadGeometry` (independent from `autoGridN`).
- I have enabled reading the lattice vectors from a XYZ file in `ppafm/cli/gui/ppafm_gui.py`, here: https://github.com/Probe-Particle/ppafm/blob/052ad8ed6dd5f686557f0895718d65cf479c7dca/ppafm/cli/gui/ppafm_gui.py#L580-L581 I am not using the GUI myself and did not test it. Please check that I have not done anything wrong by this.
- When testing my code on the example in `examples/PTCDA_single`, one with a tilted tip, I noticed an error message printed by `plot_results.py`. That was because `plot_results.py` attempted to `del fzs`, but for a tilted tip, `fzs` is actually replaced by `Fout`. I have repaired this problem.
- In `ppafm/ocl/relax.py`, I have added the `self` argument to the definition of the `prepareFEConv` method, see here https://github.com/Probe-Particle/ppafm/blob/052ad8ed6dd5f686557f0895718d65cf479c7dca/ppafm/ocl/relax.py#L538 I was forced to do so by one of the precommit hooks (namely `mypy`), which complained that at least one argument is needed here. I have zero experience with object-oriented programming in Python so I hope I have guessed correctly it should be the `self` rather than any other argument that belongs there.
- I have put the correct spelling of 'Lennard-Jones' wherever the double-N in the guy's name was missing. The LJ potential is named after John Lennard-Jones, not Philipp Lenard. I think it's fair to keep the distinction. Nothing to do with the main purpose of this pull request, I just stumbled over it when reviewing the code to see how to integrate my contribution. **Impotant notice:** You need to recompile the C++ parts (by calling `make` in the `ppafm/cpp/` directory) when you first merge this change into you code, as `getLennardJonesFF` is a name of one of the C-implemented functions.
- Finally, I have accepted all the automatic format changes done by the precommit hooks in all the files I have touched. I somewhat regret doing it now. At least, I should have allowed the changes only as a separate commit. This way, it became nigh impossible to review my own changes with `git diff`, or even `git diff -w`. But it has been done, it needed to be done sooner or later anyway, and it cannot be undone easily now, so let it be as it is.

mondracek:
Merging this branch will also close #201. I have put the change required to solve this latter issue essentially by mistake here but maybe it is good that I did so after all. Both issues #52 and #201 required changes in the `loadGeometry` function and merging the corresponding changes from separate branches later could be problematic.

mondracek:
I have a problem now, however. Merging is blocked by some automatic tests, namely "Test docker image on ubuntu-latest". I do not even know what exactly the test does, I do not understand the corresponding error message, which says
```
ERROR: failed to solve: process "/bin/sh -c apt-get update --yes &&                           apt-get install --yes --no-install-recommends     build-essential                                   python3                                           python-is-python3                                 wget &&                                           apt-get clean && rm -rf /var/lib/apt/lists/*" did not complete successfully: exit code: 100
```
and I definitely do not think this message has anything to do with what I have done. I do not like this happening to me. Testing our work is important, but I want *my* work being tested, not some compatibility issues between the latest Ubuntu OS and the latest Python 3 or whatever this is.

NikoOinonen:
> I have a problem now, however. Merging is blocked by some automatic tests, namely "Test docker image on ubuntu-latest"

It looks to me like it's failing when building the Docker image. Could be just some temporary glitch in the Ubuntu repositories. I would ignore it for this PR, because it is not related.

@yakutovicha maybe you can have a look, if you have some idea what could be going wrong.

mondracek:
Once we make the merging work again, I suggest we merge this PR as **last** of all the hanging pull requests. Because of the formatting changes I have (unfortunately) allowed in one commit together with the substantial changes, there could arise problems merging the other branches if we attempt to merge them *after* this one.

NikoOinonen:
@mondracek This one also has a lot of formatting changes. Would it be possible for you to revert the formatting changes here as well to make it easier to see what actually changed?

mondracek:
@NikoOinonen, I was about to reply "no" to your question whether the formatting changes can be removed, but then, I realized I am not able to identify all the important changes in this branch myself any more. So I will have to do it somehow, the current state makes any review virtually impossible. Do you have an idea how to do it in a manageable way? Perhaps doing all the formatting changes in a separate branch and then rebasing this branch on it? Otherwise, I will have to redo the branch essentially from scratch.

NikoOinonen:
@mondracek 
> Do you have an idea how to do it in a manageable way? Perhaps doing all the formatting changes in a separate branch and then rebasing this branch on it? Otherwise, I will have to redo the branch essentially from scratch.

Yes, if you redo the formatting changes in another branch, then do a `git diff` between this branch and that one, that should allow you to see what changed without the formatting changes. You should be able to make a patch file with the diff and the apply it to a fresh branch. E.g. something like the following would probably do it
```
git diff one_branch other_branch > changes.patch
git checkout main
git pull
git checkout -b fresh_branch
git apply changes.patch
```

mondracek:
Replaced by #213

PR #202, opened 2023-09-22
Improve Fz2df conversion

### LINKED ISSUE #199: Wrong height of the scanned region in df.xsf
The third vector delineating the (height of) the scanning region is wrong in the `df.xsf` file generated by `plot_results.py --save_df`. The origin recorded in this file is probably wrong too, though this depends on the convention we want to adopt for the _z_ position of the oscillating tip: If we use the lower turning point of the oscillating tip as its height _z_, it would be okay, but I think we should rather use the _z_ coordinate of the central point around which the tip oscillates and then the origin is wrong too. Anyway, either the scan starts at `params['scanMin'] + A/2` and finishes at `params['scanMax'] - A/2`, or it starts at `params['scanMin']` and finishes at `params['scanMax'] - A`. In any case, the height of the scanned region must be `params['scanMax'] - params['scanMax'] - A`. What happens instead is that `plot_results.py` blindly copies `lvec` from `OutFz.xsf` to `df.xsf` without subtracting the amplitude `A` from the third box-defining vector. See here https://github.com/Probe-Particle/ppafm/blob/2344a7b01000bbf8ce3d2ddbf92abf3b9fe917c9/ppafm/cli/plot_results.py#L169-L180
This issue may be related to #6 but it happens with any kind of tip, not just the complex one. The obvious way to repair it is to modify the `lvec` array by subtracting the amplitude from its appropriate element before writing `df.xsf`. We just have to be careful about the exact value of the amplitude; as I understand it, it is always rounded to an integer number of grid points.

mondracek:
I have defined `lvec_df` in `ppafm/cli/plot_results.py` so that it contains a modified copy of `lvec`. The modification consists basically in subtracting the amplitude from the height (defined by the third vector) and adding half of the amplitude to the origin (the "zeroth" vector). This `lvec_df` is then used when saving `df.xsf` (requested by the `--save_df` option). This will close #199.
Besides that, I have improved the numerical conversion from force to frequency shift itself, as implemented by the functions `Fz2df` and `Fz2df_tilt` in `ppafm/common.py`. Well, changed at least, I hope that it is an improvement, although I made the numerical procedure more complicated. In particular, I have totally rewritten the function `getDfWeight`. My intention was to make it work for very small amplitudes. The former solution could become imprecise if the amplitude was comparable to the grid step, thus extending over only very few grid points. If the amplitude equaled the grid step or if it was even smaller, the algorithm failed completely. My proposed solution should return a frequency shift proportional to the gradient of the force (estimated from forces on two consecutive grid points) for amplitudes smaller than or equal to the grid step. Moreover, while the former solution always rounded the amplitude to some integer multiple of grid step, now the conversion result is a continuous function of the (real-typed) amplitude, even though, of course, a discrete convolution will still be done.
Not sure I would have implemented the new way of `Fz2df` if I had realized from the beginning how complicated the formulae were going to be. Nevertheless, here it is. Please decide if you like the change. It is going to impact all AFM images, though I have checked using the `Graphene/` example that the changes are very small, usually not visible by eye at all except for extremely small amplitudes. I should probably write up a description of the algorithm in LaTEX and put it up somewhere, or perhaps put it to the Wiki. So far, however, I just described it by comments in the code of the `getDfWeight` function. Unfortunately, the math rendered in the plain text format is very hard to read, I am afraid.
If you do not like the new force to `df` conversion, I will put back the old solution or push some alternative less radically different from the old one, keeping just the height correction for `df.xsf` in this branch.

### DIFF
--- a/ppafm/common.py
+++ b/ppafm/common.py
@@ -11,228 +11,232 @@ verbose = 0
 
 # ====================== constants
 
-eVA_Nm               =  16.0217657
-CoulombConst         = -14.3996448915;
+eVA_Nm = 16.0217657
+CoulombConst = -14.3996448915
 
 # default parameters of simulation
-params={
-    'PBC': True,
-    'nPBC' :       np.array( [      1,        1,        1 ] ),
-    'gridN':       np.array( [ -1,     -1,   -1   ] ).astype(int),
-    'gridA':       np.array( [ 12.798,  -7.3889,  0.00000 ] ),
-    'gridB':       np.array( [ 12.798,   7.3889,  0.00000 ] ),
-    'gridC':       np.array( [      0,        0,      5.0 ] ),
-    'FFgrid0':     np.array( [ -1.0, -1.0, -1.0 ] ),
-    'FFgridA':     np.array( [ -1.0, -1.0, -1.0 ] ),
-    'FFgridB':     np.array( [ -1.0, -1.0, -1.0 ] ),
-    'FFgridC':     np.array( [ -1.0, -1.0, -1.0 ] ),
-    'moleculeShift':  np.array( [  0.0,      0.0,    0.0 ] ),
-    'probeType':   'O',
-    'charge':      0.00,
-    'Apauli':    18.0,
-    'Bpauli':     1.0,
-    'ffModel':     'LJ',
-    'Rcore':    0.7,
-    'r0Probe'  :  np.array( [ 0.00, 0.00, 4.00] ),
-    'stiffness':  np.array( [ -1.0, -1.0, -1.0] ),
-    'klat': 0.5,
-    'krad': 20.00,
-    'tip':  's',
-    'sigma': 0.7,
-    'scanStep': np.array( [ 0.10, 0.10, 0.10 ] ),
-    'scanMin': np.array( [   0.0,     0.0,    5.0 ] ),
-    'scanMax': np.array( [  20.0,    20.0,    8.0 ] ),
-    'scanTilt': np.array( [  0.0,    0.0,   -0.1 ] ),
-    'tiltedScan': False,
-    'kCantilever'  :  1800.0,
-    'f0Cantilever' :  30300.0,
-    'Amplitude'    :  1.0,
-    'plotSliceFrom':  16,
-    'plotSliceTo'  :  22,
-    'plotSliceBy'  :  1,
-    'imageInterpolation': 'bicubic',
-    'colorscale'   : 'gray',
-    'colorscale_kpfm'   : 'seismic',
-    'ddisp'        :  0.05,
-    'aMorse'       :  -1.6,
-    'tip_base':  np.array( ['None', 0.00 ]),
-    'Rtip'         :  30.0,
-    'permit'       :  0.00552634959,
-    'vdWDampKind' : 2,
-    '#' : None
+params = {
+    "PBC": True,
+    "nPBC": np.array([1, 1, 1]),
+    "gridN": np.array([-1, -1, -1]).astype(int),
+    "gridA": np.array([12.798, -7.3889, 0.00000]),
+    "gridB": np.array([12.798, 7.3889, 0.00000]),
+    "gridC": np.array([0, 0, 5.0]),
+    "FFgrid0": np.array([-1.0, -1.0, -1.0]),
+    "FFgridA": np.array([-1.0, -1.0, -1.0]),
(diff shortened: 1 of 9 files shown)


ppafm/common.py
```diff
@@ -230,40 +230,72 @@ def add_arguments(self, arg_names):
...
+        f[1:-1] = np.sqrt(1 - t[1:-1]*t[1:-1])
+        f2[1:-1] = (np.arccos(-t[1:-1]) - t[1:-1]*f[1:-1]) / 2.0
+    f2[0] = np.pi #end point for t>=1 must be as if t=1 even if t>1
+    df = f[:-1] - f[1:] #numerical derivative (first derivative integrated over one-step-long intervals) of function f(t)
+    df2 = f2[:-1] - f2[1:] #numerical derivative (first derivative integrated over one-step-long intervals) of f2(t)
+    w[1:] = t[:-1]*df + df2 #coefficients to multiply F[i]
+    w[:-1] -= t[1:]*df + df2 #coefficients to multiply F[i+1]
+    return w/(np.pi*dz)
+
+def Fz2df( F, dz=0.1, k0 = params['kCantilever'], f0=params['f0Cantilever'], A = 0.1, units=16.0217656 ):
```
yakutovicha, line 281: I suggest changing A to the `amplitude` to be more explicit.

Also, please mention in the comments that the `amplitude` is peak-to-peak (not the usual definition). Is that correct, @ProkopHapala?

ppafm/common.py
```diff
@@ -230,40 +230,72 @@ def add_arguments(self, arg_names):
 # ============================== Pure python functions
 # ==============================
 
-def getDfWeight( n, dz=0.1 ):
+def getDfWeight( A, dz = 0.1):
```
yakutovicha, line 233: Please rename it to `get_df_weights`

yakutovicha (commented):
Thanks, @mondracek. Please add simple tets to the methods from `common.py` you are updating.

ppafm/common.py
mondracek: Yes, it is peak-to-peak. I would actually change it if were just up to me, but that would be a too hard blow to backward compatibility for others I guess.

ppafm/common.py
```diff
@@ -230,72 +230,111 @@ def add_arguments(self, arg_names):
...
+    #The largest index of w[i] corresponds to the smallest index i of force F[j+i].
     #In contrast to the theoretical description above, we already reverse all the auxiliary arrays (t,f,f2,df,df2) in the code,
     #starting from t[n+1-i] -> t[i], so that we do not need to reverse w explicitely in the end.
-    t = np.linspace(-1 + 2*n*dz/A, -1, n+1)
+    t = np.linspace(-1 + 2*n*dz/Amp, -1, n+1)
     if(n>1):
         f[1:-1] = np.sqrt(1 - t[1:-1]*t[1:-1])
         f2[1:-1] = (np.arccos(-t[1:-1]) - t[1:-1]*f[1:-1]) / 2.0
-    f2[0] = np.pi #end point for t>=1 must be as if t=1 even if t>1
+    f2[0] = np.pi/2 #end point for t>=1 must be as if t=1 even if t>1
```
mondracek: `pi/2` belongs here, not just `pi`.

mondracek:
To sum up what should be still considered for review in this PR:

1. Should we keep both **`get_df_weight` and `get_simple_df_weight`**? I am for keeping them but maybe I am too attached to the work I have done. I am open to the argument that one of them be eliminated for the sake of simplicity, if you feel so.
2. The **`Amp` vs `amplitude`** naming controversy. I prefer to keep the names as  they are now. If anyone prefers a change, please pull this branch, make the change, look at the result, decide whether you think it makes the code more legible, and if you are happy with it, push it back here. As long as it is just renaming and does not break any functionality,  I will approve it.

mondracek:
@yakutovicha or anyone else, is there anything that needs to be done with this PR yet? If not, I would like to merge it.

ProkopHapala:
> @yakutovicha or anyone else, is there anything that needs to be done with this PR yet? If not, I would like to merge it.

please go ahead Martin

mondracek:
Okay then, @ProkopHapala and @yakutovicha , please give me a formal approval (click on "Add your review", go to "Review changes", select "Approve" and confirm by clicking on "Submit review").

yakutovicha (approved):
I approve, @mondracek, thanks a lot for your work!

PR #207, opened 2023-09-28
Sign of atomic forces in the C++ version

### LINKED ISSUE #149: Odd sign convention in force calculation
When I was implementing the vdw damping functions in the OpenCL code (#130), I was kind of mystified why all of the signs for the forces were reversed in the formulas compared to the C++ implementation. For example, in the case with no damping, in C++ we have https://github.com/Probe-Particle/ppafm/blob/8763be1e7e851cb47a1294d49eb04fcd1c14b34b/ppafm/cpp/Forces.h#L138-L141 and in OpenCL we have https://github.com/Probe-Particle/ppafm/blob/8763be1e7e851cb47a1294d49eb04fcd1c14b34b/ppafm/ocl/cl/FF.cl#L583-L587 Aside from the different variable names, the only difference is one minus sign on the force vector. However, the result seems to come out the same.

Doing the derivative `F = -dE/dr` on paper, it looks to me like the formula in the OpenCL code is the right one. And it looks to me there is another sign error in the C++ code here https://github.com/Probe-Particle/ppafm/blob/8763be1e7e851cb47a1294d49eb04fcd1c14b34b/ppafm/cpp/ProbeParticle.cpp#L225 where we are calling the force/energy evaluation function with the vector between the probe and the atoms. It should be `rProbe-Ratoms[i]` instead. 

To see this, consider that when we evaluate the force, we are always end up taking the derivative of the distance between the probe and the sample atom, which is something like $\left| a - b \right|$. The derivative of this expression with respect to $a$ is $(a - b) / \left| a - b \right|$. Notice that $a$ is on the front. If we took the derivative with respect to $b$ instead, it would be $(b - a) / \left| a - b \right|$, with the sign reversed and $b$ in the front. In the probe particle model we are always calculating the force on the probe particle, so the derivative should also be with respect to the probe particle position, so it should appear in the front with positive sign.

I'm not sure if this is worth fixing since the result is correct, but I wanted to document it somewhere, because it was causing some headache for me.

@ProkopHapala @ondrejkrejci

mondracek:
Fixes #149

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.10
+FROM ubuntu:22.04
 
 # build-essential is needed for building the C++ code in the PPAFM package
 
--- a/ppafm/common.py
+++ b/ppafm/common.py
@@ -11,8 +11,8 @@ verbose = 0
 
 # ====================== constants
 
-eVA_Nm               =  16.0217657
-CoulombConst         = -14.3996448915;
+eVA_Nm               =  16.021766
+CoulombConst         =  14.399645;
 
 # default parameters of simulation
 params={
--- a/ppafm/cpp/Forces.h
+++ b/ppafm/cpp/Forces.h
@@ -6,10 +6,7 @@
 #include "Vec2.h"
 #include "Vec3.h"
 
-#define COULOMB_CONST  14.3996448915f
-
-//const double kcoulomb   = 14.3996448915;
-//const double R2SAFE     = 1.0e-8;
+#define COULOMB_CONST  14.399645f
 
 #define RSAFE   1.0e-4f
 #define R2SAFE  1.0e-8f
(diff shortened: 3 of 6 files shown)


NikoOinonen:
@mondracek I cannot tell what changed in `relaxed_scan.py` because there are so many formatting changes. We are currently doing the formatting fixes in separate PRs, so would it be possible for you to redo the changes without the black formatter?

mondracek:
@NikoOinonen I have deleted all but the first commit in this branch and redone all the substantial changes from the other commits without formatting, as an amend to the first commit. 
There should have been no changes apart from formatting in `relax.py`, by the way, so now that file is completely unchanged.

NikoOinonen (approved):
Everything looks good to me. Thanks for fixing this.

PR #209, opened 2023-10-05
Fixed incorrect loading of aims .in and poscar files.

### LINKED ISSUE #201: ppafm.io.loadGeometry returns wrong properties with FHI-aims format input
When calculating the LJ potential with e.g. `ppafm-generate-ljff` using input in FHI-aims format, io.loadGeometry should return `atoms, nDim, lvec`. See line 149 here:
https://github.com/Probe-Particle/ppafm/blob/2344a7b01000bbf8ce3d2ddbf92abf3b9fe917c9/ppafm/HighLevel.py#L143-L149

io.loadGeometry() further calls io.loadGeometryIN(fname) which should return `atoms, nDim, lvec` but instead it returns `atoms, Zs, lvec`

https://github.com/Probe-Particle/ppafm/blob/2344a7b01000bbf8ce3d2ddbf92abf3b9fe917c9/ppafm/io.py#L117-L138

~~Should we read gridN from params.ini to get the dimensions, or assume something like a 0.1Å grid?~~

E: I see there was some previous discussion at least in #197 and #52 on how to handle the grid so it is probably not needed here. The bug with the return value is still relevant.

NikoOinonen:
Fixes #201 

Fixed a problem where `loadGeometry` in `ppafm/io.py` was assuming incorrect return format for the function `loadGeometryIN` which loads FHI-aims geometry files. The same problem was also true for the `loadPOSCAR` function that loads POSCAR file formats.

### DIFF
--- a/ppafm/io.py
+++ b/ppafm/io.py
@@ -358,11 +358,13 @@ def loadNCUBE( fname ):
     return [ int(sth1[0]), int(sth2[0]), int(sth3[0]) ]
 
 def loadGeometry(fname=None,format=None,params=None):
+
     if(verbose>0): print("loadGeometry ", fname)
     if fname == None:
         raise ValueError("Please provide the name of the file with coordinates")
     if params == None:
         raise ValueError("Please provide the parameters dictionary here")
+
     if format == None or format == "":
         if fname.lower().endswith(".xyz"):
             format = "xyz"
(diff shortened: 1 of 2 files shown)


mondracek (approved):
Looks good, @NikoOinonen. It is actually a duplicate of my commit ae79b820003dac1c37022cbb80a23384b9b4161f in PR #197, but your solution is slightly better and you also created the test for it. Moreover, #197 will have to be redone because of the unfortunate formatting changes. So I agree we should merge this PR.

PR #213, opened 2023-10-23
Automatic lattice vectors

### LINKED ISSUE #52: Handling missing lattice vectors

mondracek:
Closes #52 and makes PR #197 obsolete. Formatting changes were done in the first commit. Then, a copy of the branch proposed under PR #197 was rebased on the reformatted branch. As a result, the substantial changes are in other commits than the formatting.


mondracek:
Hint for reviewers: You can use the tab "Files changed"  to view the changes. Then, on the left-hand side of the bar just below these tabs, you see the option "Changes from **all commits**". Click on it and select only the commits you want to see. I recommend the range that covers all the commits but the first one, that is, 6 commits from 50c9fc75a2630e9708fc4775495adff1dd8d1dc6 to f33562c7903b87f2030c226b06bbb442ed20cc51.

PR #216, opened 2023-10-24
Generate automatic lattice vectors in loadGeometry

### LINKED ISSUE #52: Handling missing lattice vectors

mondracek:
Fixes #52
Supersedes #213 and #197

### DIFF
--- a/ppafm/io.py
+++ b/ppafm/io.py
@@ -434,14 +434,14 @@ def loadGeometry(fname=None,format=None,params=None):
     lvec = np.array(lvec)
 
     # Shift scanning coordinates to such that actually make sense (being directly comparable to coordinates of atoms)
-    probeMin = params["scanMin"].copy()
-    probeMax = params["scanMax"].copy()
-    probeMin[0] += params["r0Probe"][0]
-    probeMin[1] += params["r0Probe"][1]
-    probeMin[2] -= params["r0Probe"][2]
-    probeMax[0] += params["r0Probe"][0]
-    probeMax[1] += params["r0Probe"][1]
-    probeMax[2] -= params["r0Probe"][2]
+    probe_min = params["scanMin"].copy()
+    probe_max = params["scanMax"].copy()
+    probe_min[0] += params["r0Probe"][0]
+    probe_min[1] += params["r0Probe"][1]
+    probe_min[2] -= params["r0Probe"][2]
+    probe_max[0] += params["r0Probe"][0]
+    probe_max[1] += params["r0Probe"][1]
+    probe_max[2] -= params["r0Probe"][2]
 
     # Generate automatic lattice vectors and grid dimensions if needed
     pad = 3.0
@@ -451,7 +451,8 @@ def loadGeometry(fname=None,format=None,params=None):
         # The automatic generated lattice vector should enclose the whole area filled with atoms as well as the whole scanning area, plus the default padding.
         if np.allclose(lvec[i + 1, :], 0):
             params["PBC"] = False
-            lvec[i + 1, i] = probeMax[i] + pad
+            #lvec[i + 1, i] = probe_max[i] - probe_min[i] + 2*pad
+            lvec[i + 1, i] = probe_max[i] + pad
 
         # Generate automatic grid using the default step if the grid dimension specified so far is not a positive number
         if not nDim[i] > 0:
--- a/ppafm/io.py
+++ b/ppafm/io.py
@@ -434,12 +434,14 @@ def loadGeometry(fname=None,format=None,params=None):
     lvec = np.array(lvec)
 
     # Shift scanning coordinates to such that actually make sense (being directly comparable to coordinates of atoms)
-    params["scanMin"][0] += params["r0Probe"][0]
-    params["scanMin"][1] += params["r0Probe"][1]
-    params["scanMin"][2] -= params["r0Probe"][2]
-    params["scanMax"][0] += params["r0Probe"][0]
-    params["scanMax"][1] += params["r0Probe"][1]
-    params["scanMax"][2] -= params["r0Probe"][2]
+    probeMin = params["scanMin"].copy()
+    probeMax = params["scanMax"].copy()
+    probeMin[0] += params["r0Probe"][0]
+    probeMin[1] += params["r0Probe"][1]
+    probeMin[2] -= params["r0Probe"][2]
+    probeMax[0] += params["r0Probe"][0]
+    probeMax[1] += params["r0Probe"][1]
+    probeMax[2] -= params["r0Probe"][2]
 
     # Generate automatic lattice vectors and grid dimensions if needed
     pad = 3.0
@@ -448,26 +450,12 @@ def loadGeometry(fname=None,format=None,params=None):
         # Zero lattice vector is considered undefined and triggers creation of an automatic one.
         # The automatic generated lattice vector should enclose the whole area filled with atoms as well as the whole scanning area, plus the default padding.
         if np.allclose(lvec[i + 1, :], 0):
-            if params["PBC"]:
-                lvec[i + 1, i] = (
-                    max(max(atoms[i + 1]), params["scanMax"][i])
-                    - min(min(atoms[i + 1]), params["scanMin"][i])
-                    + 2 * pad
-                )
-            else:
-                lvec[i + 1, i] = max(max(atoms[i + 1]), params["scanMax"][i]) + pad
+            params["PBC"] = False
+            lvec[i + 1, i] = probeMax[i] + pad
 
         # Generate automatic grid using the default step if the grid dimension specified so far is not a positive number
         if not nDim[i] > 0:
             nDim[i] = round(np.linalg.norm(lvec[i + 1, :]) / default_grid_step)
-
-    # Shift scanning coordinates back because, unfortunately, we have to keep backward compatibility
-    params["scanMin"][0] -= params["r0Probe"][0]
-    params["scanMin"][1] -= params["r0Probe"][1]
-    params["scanMin"][2] += params["r0Probe"][2]
-    params["scanMax"][0] -= params["r0Probe"][0]
-    params["scanMax"][1] -= params["r0Probe"][1]
-    params["scanMax"][2] += params["r0Probe"][2]
 
     # Copy lattice vectors and grid dimensions to the global parameters
     # so as to guarantee compatibility between the local variables and global parameters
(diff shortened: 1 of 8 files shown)


mondracek:
The formatting changes were removed with the help of @yakutovicha and the other commits squashed into one.

ppafm/io.py
```diff
@@ -415,9 +416,68 @@ def loadGeometry(fname=None,format=None,params=None):
...
+    for i in range(3):
+        # Zero lattice vector is considered undefined and triggers creation of an automatic one.
+        # The automatic generated lattice vector should enclose the whole area filled with atoms as well as the whole scanning area, plus the default padding.
+        if np.allclose(lvec[i + 1, :], 0):
+            if params["PBC"]:
+                lvec[i + 1, i] = (
+                    max(max(atoms[i + 1]), params["scanMax"][i])
+                    - min(min(atoms[i + 1]), params["scanMin"][i])
+                    + 2 * pad
+                )
```
NikoOinonen, line 456: Do we really need to consider the atom positions here? It should be enough to just have enough padding around the scan region so that the probe particle stays inside the FF grid during the scan. Sometimes you just want a smaller scan of a bigger structure.

ppafm/io.py
```diff
@@ -415,9 +416,68 @@ def loadGeometry(fname=None,format=None,params=None):
...
+            lvec.append(params["gridC"].copy())
+    lvec = np.array(lvec)
+
+    # Shift scanning coordinates to such that actually make sense (being directly comparable to coordinates of atoms)
+    params["scanMin"][0] += params["r0Probe"][0]
+    params["scanMin"][1] += params["r0Probe"][1]
+    params["scanMin"][2] -= params["r0Probe"][2]
+    params["scanMax"][0] += params["r0Probe"][0]
+    params["scanMax"][1] += params["r0Probe"][1]
+    params["scanMax"][2] -= params["r0Probe"][2]
```
NikoOinonen, line 442: Since these are set back in the end, it would probably be clearer if you jsut make a copy of the "scanMin" and "scanMax" values here and use those in the following computation, so that you don't have to set them back in the end.

ppafm/io.py
mondracek: Yes, you are right. I will redo it.

ppafm/io.py
mondracek: Okay

mondracek:
@Niko, I implemented the amendments you proposed (not pushed yet). Restricting the automatic cell to the scan region (plus some padding) works okay for non-periodic structures. It leads to wrong results for periodic `PBC True` in `params.ini`. That would not be a problem since the automatically generated cell is intended for clusters and isolated molecules while for periodic structures, the lattice vectors should be given explicitly. However, it may become a problem because `params["PBC"]=True` is our default now. I have already proposed to make `PBC False` the default in the discussion under #52 but forgot about it later.

What I'd proposed now: Make `params["PBC"]=None` by default and unless `PBC True` or `PBC False` is specified explicitly in `params.ini`, make it `True` when reading the input from `*.xsf` or from `geometry.in` without `lattice_vectors`, and make `False` when reading from the `*.xyz` file or a `geometry.in` file without `lattice_vectors`. I am not sure about `*.cube` files, whether they should change `params["PBC"]=None` to `params["PBC"]=True` or to `params["PBC"]=False`. Probably to `false` as the CUBE format is originally intended for non-periodic systems although FHI-AIMS often uses it for periodic ones too. As for the XSF format, it could actually describe a non-periodic system too, but we are currently not making the distinction. Should I implement this default handling of `PBC` in this PR?

NikoOinonen:
> What I'd proposed now: Make params["PBC"]=None by default and unless PBC True or PBC False is specified explicitly in params.ini

Yes, it makes sense to have some default behaviour when not otherwise specified.

> make it True when reading the input from *.xsf or from geometry.in with lattice_vectors, and make it False when reading from the *.xyz file or a geometry.in file without lattice_vectors. I am not sure about *.cube files, whether they should change params["PBC"]=None to params["PBC"]=True or to params["PBC"]=False.

Would it not be fine to just set it True whenever the lattice vectors are defined in the input regardless of the format (when not explicitly set as False)? Especially for cube files (and for xsf), the convolution of the Hartree potential is always periodic just because of how the operation is defined, so we end up using periodic boundary conditions anyways.

>  Should I implement this default handling of PBC in this PR?

Yes, I think it makes sense, since it affects the behaviour here.

mondracek:
Right, I was about to do it that way after all: Set `params["PBC"]` in `loadGeometry` depending on whether the `lvec` is `None` or something else. However, I realized we must not initialize `params["PBC"]` as `None` in `common.py` as I tried to do. The type of each parameter is determined by its initial value and when it is unknown, reading the parameter from `params.ini` does not work. So I will have to leave the default value of `params["PBC"]` to be `True`. But then, I must not force it to be `False` even if e.g. the input is a *xyz file without lattice vectors because the lattice vectors can still be specified as  the `gridA`, `gridB`, `gridC` parameters in `params.ini` and the user can mean these as true lattice vectors rather than just dimensions of a non-periodic grid.
So in the end, the only thing I can do with `params["PBC"]` in response to the geometry read from the input is to put `params["PBC"]=False` exactly when the automatic grid is being generated and leave it as it was (set by default or by the user in `params.ini`) otherwise.

mondracek:
Pushed a new version that reflects the recent discussion.

ppafm/io.py
```diff
@@ -415,9 +416,56 @@ def loadGeometry(fname=None,format=None,params=None):
...
+    probeMax[0] += params["r0Probe"][0]
+    probeMax[1] += params["r0Probe"][1]
+    probeMax[2] -= params["r0Probe"][2]
+
+    # Generate automatic lattice vectors and grid dimensions if needed
+    pad = 3.0
+    default_grid_step = 0.1
+    for i in range(3):
+        # Zero lattice vector is considered undefined and triggers creation of an automatic one.
+        # The automatic generated lattice vector should enclose the whole area filled with atoms as well as the whole scanning area, plus the default padding.
```
NikoOinonen, line 451: Incorrect comment. Not with atoms anymore.

ppafm/io.py
```diff
@@ -415,9 +416,56 @@ def loadGeometry(fname=None,format=None,params=None):
...
+            lvec.append(params["gridA"].copy())
+        if len(lvec) == 2:
+            lvec.append(params["gridB"].copy())
+        if len(lvec) == 3:
+            lvec.append(params["gridC"].copy())
+    lvec = np.array(lvec)
+
+    # Shift scanning coordinates to such that actually make sense (being directly comparable to coordinates of atoms)
+    probeMin = params["scanMin"].copy()
+    probeMax = params["scanMax"].copy()
```
NikoOinonen, line 438: A minor thing, but I think we want to eventually move toward the PEP8 standard for naming conventions, where variable names are in snake_case instead of camelCase, so here it would be `probe_min` instead of `probeMin`.

ppafm/io.py
```diff
@@ -415,9 +416,56 @@ def loadGeometry(fname=None,format=None,params=None):
...
+
+    # Generate automatic lattice vectors and grid dimensions if needed
+    pad = 3.0
+    default_grid_step = 0.1
+    for i in range(3):
+        # Zero lattice vector is considered undefined and triggers creation of an automatic one.
+        # The automatic generated lattice vector should enclose the whole area filled with atoms as well as the whole scanning area, plus the default padding.
+        if np.allclose(lvec[i + 1, :], 0):
+            params["PBC"] = False
+            lvec[i + 1, i] = probeMax[i] + pad
```
NikoOinonen, line 454: It seems to me like we are not taking into account the start of the scan region here. I think we would need to adjust the origin of the lvec to be at something like `probeMin - pad`, and then the length of the lattice vectors would be something like `probeMax - probeMin + 2 * pad`? Notice that currently `probeMin` is not used.

NikoOinonen (changes requested):

ppafm/io.py
mondracek: Corrected

ppafm/io.py
```diff
@@ -451,7 +451,8 @@ def loadGeometry(fname=None,format=None,params=None):
         # The automatic generated lattice vector should enclose the whole area filled with atoms as well as the whole scanning area, plus the default padding.
         if np.allclose(lvec[i + 1, :], 0):
             params["PBC"] = False
-            lvec[i + 1, i] = probeMax[i] + pad
+            #lvec[i + 1, i] = probe_max[i] - probe_min[i] + 2*pad
```
mondracek, line 454: This comment line will be activated once we deal with a potential non-zero origin properly.

ppafm/io.py
mondracek: Exactly right. The thing is, the current version of the code (at least the CPU part, I do not know about the OCL) does not work with non-zero origin in non-periodic systems. I should and I will solve this, but as a separate issue. In fact, I have added a line ``#lvec[i + 1, i] = probe_max[i] - probe_min[i] + 2*pad`` to `io.py`, which is commented-out for now but which I intend to use instead of the contemporary ``lvec[i + 1, i] = probe_max[i] + pad`` as soon as I solve the issue of an arbitrary origin.

ppafm/io.py
mondracek: Okay, I will change it.

mondracek: Done

ppafm/io.py
NikoOinonen: Ok, I didn't know that. It should work with any origin in the OCL code. I guess this by extension means that negative scan window coordinates don't work when `PBC = False`. Is there an issue for this? If not, then better make one so that it does not slip by.

NikoOinonen (approved):
Ok, looks good now. Thanks Martin!

ppafm/io.py
mondracek: Exactly, last time I tried I got a blank area with zero forces for the part of scan region in which any of the three coordinates negative. There is no issue about it and I am going to make one. The original plan was to solve it as a by-product of #217 (may seem to be unrelated but as FHI-AIMS tends to be creative about the grid origin, it would have to be done to address that issue too). However, as I am being rather discouraged to implement #217, I will start a separate issue.

Issue #210, opened 2023-10-05
Protocol for dealing with formatting changes needed


mondracek:
We need to establish some recommended procedure how to deal with formatting changes required by the pre-commit hooks like `flake` and `black`. I am now being forced to redo part of my work on the code because my changes got buried in a flood of the automatic formatting changes I have foolishly accepted and no one, myself included, can tell apart the actual changes in the code submitted for review.
Should we perhaps make a separate branch in which the automatic formatting will be applied to all files and nothing else will be done there? Then merge it into main and base all future work on the newly formatted files?

NikoOinonen:
Yes, we should get on with this.

@yakutovicha We did one PR, but then stopped. What is your status on this?

In the meanwhile, I found that if you do `SKIP=black commit` then it does not apply black during the pre-commit hook.

yakutovicha:
Prokop will select parts of the code that shouldn't be formatted, and then we will apply black to the whole repository.

PR #219, opened 2023-10-27
Formatting

NikoOinonen:
Added `# fmt: off/on` according to discussion in #210. Also applied black to a couple of the files.

Also, I found it useful to run black from inside VScode to see immediately what it would look like, so I added the
```toml
[tool.black]
line-length = 180
```
to `pyproject.toml`, which allows black inside the editor to automatically pick up the correct line length.


NikoOinonen:
With this I'm happy if we just run black over the whole repo. I'll leave it to @ProkopHapala to check if he has any more things that need manual formatting.

yakutovicha (approved):
LGTM!

PR #222, opened 2023-10-31
Apply `pre-commit` for the whole repository

yakutovicha:


yakutovicha:
The only problem with the current setup is that I cannot exclude the `ppafm/dev` folder (I tried different ways).

So if one tries to run `pre-commit run --all-files`, then everything is modified. I manually removed all modifications from the `dev` folder to make this PR.

NikoOinonen (approved):
I didn't look through everything in detail, but since it's just formatting changes, should be fine.

We should also remember to document the `# fmt: on/off` in the wiki, so maybe don't close #210 yet, or make a new issue for it.

yakutovicha:
> We should also remember to document the `# fmt: on/off` in the wiki, so maybe don't close #210 yet, or make a new issue for it.

I just [did it](https://github.com/Probe-Particle/ppafm/wiki/For-Developers#disable-auto-formatting), as it was quicker than making an issue.

Issue #118, opened 2023-03-21
Enable smoke tests. 


yakutovicha:
We might adapt the existing examples for that purpose. We should configure pytest to look at the examples and make sure they run successfully. The examples folder should remain in the root folder.

PR #228, opened 2023-11-10
Run PTCDA_single example with pytest

yakutovicha:
fixes https://github.com/Probe-Particle/ppafm/issues/118
supersedes #205


yakutovicha:
Ok, with this, I declare a partial win against tests. To achieve it, I had to do two things:

* Disable `test_vdw.py` by renaming it to `_test_vdw.py`
* Disable tests for Python 3.8.

It looks to me that `test_vdw.py` interferes with the actual smoke tests through the `common.py` module that stores the current value of parameters. Therefore, it becomes urgent to fix https://github.com/Probe-Particle/ppafm/issues/232.

For the Python 3.8 issue, I don't know the origin of the problem.

yakutovicha:
For your information, this is the current test coverage:

```
---------- coverage: platform linux, python 3.9.18-final-0 -----------
Name                                                  Stmts   Miss  Cover
-------------------------------------------------------------------------
examples/PTCDA_Hartree_dz2/example_ptcda_hartree.py      18      1    94%
examples/PTCDA_single/example_ptcda.py                   15      1    93%
ppafm/GridUtils.py                                      100     56    44%
ppafm/HighLevel.py                                      320    129    60%
ppafm/PPPlot.py                                         155    121    22%
ppafm/__init__.py                                         2      0   100%
ppafm/atomicUtils.py                                    203    180    11%
ppafm/cli/__init__.py                                     0      0   100%
ppafm/cli/generateElFF.py                               112     71    37%
ppafm/cli/generateElFF_point_charges.py                  11      1    91%
ppafm/cli/generateLJFF.py                                16      3    81%
ppafm/cli/plot_results.py                               174     73    58%
ppafm/cli/relaxed_scan.py                               138     47    66%
ppafm/common.py                                         500    267    47%
ppafm/core.py                                           217     69    68%
ppafm/cpp_utils.py                                       53     23    57%
ppafm/defaults/__init__.py                                0      0   100%
ppafm/defaults/d3.py                                     29     14    52%
ppafm/elements.py                                         7      0   100%
ppafm/fieldFFT.py                                       317    170    46%
ppafm/io.py                                             621    294    53%
ppafm/ocl/AFMulator.py                                  361    178    51%
ppafm/ocl/__init__.py                                     0      0   100%
ppafm/ocl/field.py                                     1078    773    28%
ppafm/ocl/oclUtils.py                                    64     32    50%
ppafm/ocl/relax.py                                      343    232    32%
ppafm/version.py                                          1      0   100%
tests/human_eye/test_TipForce.py                         29      0   100%
tests/human_eye/test_splines.py                          26      0   100%
tests/human_eye/test_vdwDamping.py                       67      1    99%
tests/test_afmulator.py                                  27      0   100%
tests/test_atomicUtils.py                                 5      0   100%
tests/test_common.py                                     12      0   100%
tests/test_datagrid.py                                   47      0   100%
tests/test_io.py                                         54      0   100%
-------------------------------------------------------------------------
TOTAL                                                  5122   2736    47%
```

47 % is not very good, but not catastrophic either.

NikoOinonen (commented):
@yakutovicha Do you think it would make more sense to put the `example_*.py` scripts under the `tests` folder? They don't really add any value as examples for the end user, so they are a kind of clutter. If you put them under `tests` then we also wouldn't need to put the underscore in any of the scripts under the `examples` folder.

examples/PTCDA_Hartree_dz2/example_ptcda_hartree.py
```diff
@@ -0,0 +1,27 @@
...
+    subprocess.run(["unzip", "LOCPOT.xsf.zip"])
+
+    generate_ljff(["--input", "LOCPOT.xsf"])
+    generate_elff(["--input", "LOCPOT.xsf", "--tip", "dz2"])
+    relaxed_scan(["--klat", "0.5", "--charge", "-0.10"])
+    plot_results(["--klat", "0.5", "--charge", "-0.10", "--Amplitude", "0.5", "--df"])
+
+
+if __name__ == "__main__":
+    example_ptcda_singe()
```
NikoOinonen, line 27: Wrong function, `example_ptcda_singe()` is being run twice now. Also `singe` -> `single`.

NikoOinonen (changes requested):

NikoOinonen:
I tried running the test with Python 3.8 locally, and interestingly the segfault only happens when running all tests with pytest. If run individually (with or without pytest), the test passes, so it seems that one of the other test functions is messing up the state somehow. No clue why it's only happening on 3.8, though.

yakutovicha:
> @yakutovicha Do you think it would make more sense to put the `example_*.py` scripts under the `tests` folder? They don't really add any value as examples for the end user, so they are a kind of clutter. If you put them under `tests` then we also wouldn't need to put the underscore in any of the scripts under the `examples` folder.

The rationale behind this is the separation of unit tests from smoke tests. Also, the way we run ppafm will undergo some transformation soon.

So, I would not worry about this now. We can revisit tests in the future when we have a clearer idea of where we are going.

examples/PTCDA_Hartree_dz2/example_ptcda_hartree.py
yakutovicha: Pytest was looking for functions starting with `example_` so both `example_ptcda_hartree` and `example_ptcda_singe` were run.

Anyway, thanks for the careful review, I fixed those things in c451e7c91c1f1e65ec0fe9649e45080cc7b5b875

yakutovicha:
> I tried running the test with Python 3.8 locally, and interestingly the segfault only happens when running all tests with pytest. If run individually (with or without pytest), the test passes, so it seems that one of the other test functions is messing up the state somehow. No clue why it's only happening on 3.8, though.

I believe it has to do with the fact that Python loads modules only once (which makes sense). If the same module is used for two different simulations, it creates problems. The params variable, for instance, will be shared among those.

This makes the case clear that storing some state variables inside a module is not a good idea. That is why I created https://github.com/Probe-Particle/ppafm/issues/232.

That is also why tests are useful in general 👍

NikoOinonen (approved):
> The rationale behind this is the separation of unit tests from smoke tests.

Makes sense, but I also kind of think that the examples directory is also not the right place for them. I would maybe put them into their own subdirectory in `tests/smoke_tests` or something, but it's not that important, so I will leave it up to you.

Issue #198, opened 2023-09-11
Suggest cleanup of geometry handling in the CPU version


mondracek:
As I mentioned in my comment to #52, the way the CLI version of our code handles the geometry parameters (`params['gridA'], ... params['gridN']`, `lvec`, `nDim` etc.) could use some cleanup. For starters, I propose we discuss this issue:
1. What do we expect to be the role of `gridA`, `gridB`, `gridC`, and `gridN` parameters in the code? Do we want them to be just an optional input (in case the geometry is not specified by other means, in particular in a `*.xsf` or other input file)? In such a case, they can be safely forgotten (I do not mean literally deleted, just ignored) after the initialization stage. The actual lattice vectors and grid size should be consistently carried in variables like `lvec` and `nDim` throughout the actual computation. Alternatively, do we want `parms['gridA'], ... params'['gridN']` to serve as a global variable? Then, we should consistently work with those. In the most extreme version of this choice, we would not even need to introduce `lvec` and `nDim` variables at all. I can somehow understand that a compromise might be desirable: Perhaps we want some functions to be also usable as stand-alone entities, without being called from the main scripts. A moderately experienced user might want to do something in an interactive Python session. The `loadGeometry` function is one of those we may want to be able to call this way. Then, it makes sense for such a function to take `lvec`, 'nDim` etc. as arguments and store them as its local variables. On the other hand, some other functions may be intended as mere subprograms of the main executable scripts and need access to global variables, since passing everything they need through arguments to them would be inefficient and would generate too long argument lists. I suspect this compromise was what @ProkopHapala had in mind when originally writing the code. Prokop, is that so? I am actually in favor of the compromise solution, given that we decide *very carefully* which functions should be the independent entities, taking geometry parameters as their arguments and perhaps returning them as their values, which functions should in contrast access the global variables, and why the chosen option is needed in every case. Because if we are not clear about this, we inevitably end up with the mess we currently have.
2. As a second point, I would like to make sure that the code reads the origin of the unit cell from the `*.xsf` and `*.cube` files every time, stores it and handles accordingly. That is, the code should not rely on the origin being zero. We might even introduce an optional parameter `params['gridOrigin']` to `params.ini`, which would be used with `*.xyz` files. That should solve the problem with negative coordinates of atoms in non-periodic systems, which is currently effectively forbidden.

mondracek:
[...]

Issue #190, opened 2023-08-07
Enable full input from Cube files, including densities


mondracek:
I would like to introduce the option to have input from ***Cube*** files essentially everywhere where input from *XSF* is possible. The reason is mainly that Cube is the standard output format for 3D maps (potentials and densities) from FHI-AIMS. While FHI-AIMS has the option to switch the output into XSF too, this has problems of its own. As pointed out by @NikoOinonen during the discussion of PR #179 , Cube cannot be used for densities now, because there is an issue with a scaling factor. Moreover, I know about an issue with how an origin of the coordinate system is chosen, which is treated in our code differently for Cube as compared to XSF. This latter problem, the one with the coordinates, does not make the results literally wrong (probably), but it is a pain to identify positions in the resulting images with the original positions of atoms in the input atomic structure. These issues are quite important to me as I have been mostly using FHI-AIMS for DFT recently and the imperfect support of the Cube format by `ppafm` really hampers my work.

NikoOinonen:
@mondracek Are you working on this? It would also make my life easier if I could load electron densities from cube files. I could also try to fix this if you are busy.

NikoOinonen:
Which issues specifically is this blocked by?

aureliojgc:
Hi guys!
I have tried to use FHI-AIMS cube files in the past for density overlaps (If I understood thats what Martin is trying) and I got a lot of noise (like an oscillation if you look at df profiles along the z axis). My best guess was that, because at the center of the atoms in AIMS there is no density (zero) it looks like a bunch on widen dirac deltas that become oscillations under the fourier transform of the integral (If I remember right, it was quite ago). We though on fit this area out since it is not necessary all that much, but never got to actually do it.
Just in case any of this info could come useful

PR #229, opened 2023-11-10
Consider non-zero origin of the grid cell

### LINKED ISSUE #223: Atoms with negative coordinates in non-periodic systems invisible (in the CPU version)
The problem can be demonstrated on `ppafm/examples/benzeneBrCl2`. While the original example works because `params.ini` sets `PBC True` to ensure artificial periodicity, the problem pops up once we set `PBC False`.
Compare the generated PNG files that show the frequency shift for one particular height above the molecule in the original setup, see here
df_015
and after changing the `PBC` option to `False` in `params.ini` here:
df_015

I stumbled over this issue when working on #216, see my discussion with @NikoOinonen over there. As far as I could tell from the source code, the issue is related to an inconsistent treatment of the grid's ***origin***. At some point, the origin is made zero (so the coordinates of all grid points are always non-negative).

Not a problem for periodic systems, e.g. a coordinate _-x_ rolls to _a<sub>x</sub> - x_, where _a<sub>x</sub>_ is the _x_ coordinate of a lattice vector. Also, the problem only appears for potentials generated from atomic positions like Lennard-Jones or point charges, not for DFT-calculated potentials and densities.

However, I should note that while the images calculated for periodic DFT systems are essentially correct even for an arbitrary grid origin in the CUBE or XSF file, the are ***shifted*** by the vector that starts at the original zero and ends at the non-zero origin (an artifact of the origin being effectively redefined to zero). Not a problem with VASP that does not define any origin of the grids outside zero anyway, but a big pain with FHI-AIMS.

mondracek:
Fixes #223. Will help with #190 too.

I have introduced a **new parameter**, `gridO`, through which the user can specify the grid cell origin in `params.ini` if the geometry input file is not `*.xsf`, `*.cube` or `*.npz`, that is, if the geometry input file itself does not specify the grid.
- You are welcome to propose a different name for this `gridO` parameter.
- Moreover, we may want to merge it with `FFgrid0`. Now `gridO` is intended for the CPU(CLI) version while `FFgrid0` is for the GPU(OCL) version. However, merging those parameters into one may require more profound changes in the CPU version, in particular, maintaining a consistent distinction between the lattice vectors of the structural periodicity and the grid vectors, as discussed in #198.

### DIFF
--- a/ppafm/HighLevel.py
+++ b/ppafm/HighLevel.py
@@ -40,6 +40,24 @@ def trjByDir(n, d=[0.0, 0.0, PPU.params["scanStep"][2]], p0=[0, 0, PPU.params["s
     return trj
 
 
+def shift_positions(R, s):
+    """
+    Shifts positions in R by s; returns the result. Needed especially to shift atoms according to the grid origin.
+
+    Arguments:
+    R = list of positional vectors or a 2D array. The first index donoting the item (typically an atom)
+      the second index determines the coordinate
+    s = vector that determines the required shift
+
+    Returns Rs: Rs[ia,i] = R[ia,i] - s[i]
+    """
+    Rs = np.array(R).copy()
+    Rs[:,0] += s[0]
+    Rs[:,1] += s[1]
+    Rs[:,2] += s[2]
+    return Rs
+
+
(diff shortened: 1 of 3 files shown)


NikoOinonen (approved):
LGTM!

PR #237, opened 2023-11-14
repair in plot_results.py for IETS (still not fully working).

ondrejkrejci:
There was one problem in plot_results.py, which is solved with this. Allows to run at least one IETS example, that we have on PPSTM.

The small change in the fastmath_light.h enables easier compilation from source on more platforms (at least it is working for me on my crazy Win machine without sudo).

And I have learned how the new formating and precommit things are working 👍 .

### DIFF
--- a/ppafm/cli/plot_results.py
+++ b/ppafm/cli/plot_results.py
@@ -150,7 +150,7 @@ def main(argv=None):
                 )
 
             if opt_dict["iets"] is not None:
-                eigenvalue_k, lvec, _ = io.load_vec_field(dirname + "/eigvalKs", data_format=args.output_format)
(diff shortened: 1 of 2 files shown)


NikoOinonen (approved):
LGTM!

NikoOinonen:
@yakutovicha Is it really the case that if you add a single line that is not tested, then it fails the CI test completely? I think this is maybe too strict of a requirement for every PR.

yakutovicha:
> @yakutovicha Is it really the case that if you add a single line that is not tested, then it fails the CI test completely? I think this is maybe too strict of a requirement for every PR.

There should be a setting for that. Sorry, I can't recall exactly how to change it. Feel free to update it.

yakutovicha:
@NikoOinonen, I think I changed the threshold to 2%. Let's hope it will work.

PR #239, opened 2023-11-28
Electric field scaling corrected for non-zero grid origin

### LINKED ISSUE #238: Wrong scaling of electric field with tip distance for an arbitrary tip origin
There was still a mistake in my implementation of arbitrary origin for the grid, #229. It concerbed the `zpos` variable defined in `ppafm/cli/generateElFF.py`. The KPFM simulation was affected when starting with electrostatic potentials in CUBE files. I am now linking a pull request that should solve the issue.

mondracek:
Fixes #238.

### DIFF
--- a/ppafm/cli/generateElFF.py
+++ b/ppafm/cli/generateElFF.py
@@ -162,8 +162,12 @@ def main(argv=None):
         ff_kpfm_tvs0, _ = computeElFF(v_v0_aux2, lvec, n_dim, drho_kpfm, computeVpot=args.energy, tilt=args.tilt, sigma=sigma)
 
         print("Linear E to V")
-        zpos = np.linspace(lvec[0, 2] - args.z0, lvec[3, 2] - args.z0, n_dim[0])
+        zpos = np.linspace(lvec[0, 2] - args.z0, lvec[0, 2] + lvec[3, 2] - args.z0, n_dim[0])
         for i in range(n_dim[0]):
+            if zpos[i] <= 0:
+                # z position of the KPFM tip with respect to the sample must not be zero or negative
+                # Should that happen, use periodicity in z to get zpos>0
+                zpos[i] += lvec[3, 2]
             ff_kpfm_t0sv[i, :, :] = ff_kpfm_t0sv[i, :, :] / ((v_ref_s) * (zpos[i] + 0.1))
             ff_kpfm_tvs0[i, :, :] = ff_kpfm_tvs0[i, :, :] / ((v_ref_t) * (zpos[i] + 0.1))
 
--- a/ppafm/cli/generateElFF.py
+++ b/ppafm/cli/generateElFF.py
@@ -164,10 +164,10 @@ def main(argv=None):
         print("Linear E to V")
         zpos = np.linspace(lvec[0, 2] - args.z0, lvec[0, 2] + lvec[3, 2] - args.z0, n_dim[0])
         for i in range(n_dim[0]):
-            if zpos[i] <= 0:
-                # z position of the KPFM tip with respect to the sample must not be zero or negative
-                # Should that happen, use periodicity in z to get zpos>0
-                zpos[i] += lvec[3, 2]
+            # z position of the KPFM tip with respect to the sample must not be zero or negative
+            # Should that happen, use periodicity in z to get zpos>0
+            zpos[i] %= lvec[3, 2]
+
             ff_kpfm_t0sv[i, :, :] = ff_kpfm_t0sv[i, :, :] / ((v_ref_s) * (zpos[i] + 0.1))
             ff_kpfm_tvs0[i, :, :] = ff_kpfm_tvs0[i, :, :] / ((v_ref_t) * (zpos[i] + 0.1))
 
--- a/ppafm/PPPlot.py
+++ b/ppafm/PPPlot.py
@@ -104,6 +104,7 @@ def plotImages(
         else:
             plt.title(r"Tip_z = %2.2f $\AA$" % zs[i])
         plt.savefig(prefix + "_%3.3i.png" % i, bbox_inches="tight")
+        # plt.savefig(prefix + "_%3.3i.svg" % i, bbox_inches="tight")
         plt.close()
 
 


mondracek:
Automatic checks failed on code coverage by tests (for the obvious reason that there are indeed no checks for this part of code yet). Can I somehow deactivate the check?

ppafm/cli/generateElFF.py
```diff
@@ -162,8 +162,12 @@ def main(argv=None):
         ff_kpfm_tvs0, _ = computeElFF(v_v0_aux2, lvec, n_dim, drho_kpfm, computeVpot=args.energy, tilt=args.tilt, sigma=sigma)
 
         print("Linear E to V")
-        zpos = np.linspace(lvec[0, 2] - args.z0, lvec[3, 2] - args.z0, n_dim[0])
+        zpos = np.linspace(lvec[0, 2] - args.z0, lvec[0, 2] + lvec[3, 2] - args.z0, n_dim[0])
         for i in range(n_dim[0]):
+            if zpos[i] <= 0:
+                # z position of the KPFM tip with respect to the sample must not be zero or negative
+                # Should that happen, use periodicity in z to get zpos>0
+                zpos[i] += lvec[3, 2]
```
NikoOinonen, line 170: I'm not familiar with the KPFM calculation, but this looks a bit odd to me. If you add one lattice vector, wouldn't it just go back to the exact same place in the periodic image? Or if it's not periodic, then it's just shifted somewhere far away. Or maybe that is exactly what you want?

NikoOinonen:
> Automatic checks failed on code coverage by tests (for the obvious reason that there are indeed no checks for this part of code yet). Can I somehow deactivate the check?

I was under the impression that we already lifted this restriction.

@yakutovicha Do you know?

mondracek:
Perhaps the restriction concerning the code coverage has not been lifted completely but rather just alleviated to some less strict percentage? Then it would not help here because the coverage of the three lines this PR modifies is exactly zero. Merging is blocked now; is an approving review enough to unblock it, even with the failed test?

NikoOinonen:
> Merging is blocked now; is an approving review enough to unblock it, even with the failed test?

Yes, should be. And even the review thing can be bypassed if need be.

ppafm/cli/generateElFF.py
mondracek: Perhaps a shorter answer: The value of `zpos` is not used as a kind of index to locate the tip. Rather, a division by this value is performed on a certain field, so the difference by the lattice vector matters.

NikoOinonen: Okay, just making sure that it's correct.

> in case the grid origin were somewhere absurdly negative

I think it would work if you just do a modulo operation: 
```python
zpos[i] %= lvec[3, 2]
```

aureliojgc: Hi! Sorry for the late response.
Yes the idea of zpos together with z0 is to divide the field by the distance between the metallic parts of the tip and the sample. And it cant be zero, that's why I was adding 0.1 in some lines around there.
Bu I can't recal ading a whole z vector to solve as you guys show there... So I can't really tell what was I intending there...

Anyway I am checking the changes and all looks OK for me

aureliojgc (changes requested):
I think now I see what's ging on.
1st - yes, I was not taking into acount the posibility of the origin being not zero. Correcting that is right

2nd - ading a whole vector to zpos to avoid to divide by zero will make you divide by a much larger number (besides making the results depend on the dimensions of your unit cell) making the LCPD much smaller than it should be...
I would not make that change and use an if or a while instead to avoid divisions by zero, since a singularity there was inducing noise al over if I remember properly

ppafm/cli/generateElFF.py
mondracek: Hi @aureliojgc , the thing is that not only `zpos` should be non-zero, it should be positive, as dividing by `zpos`<0 also does not make much sense. Well, it might make sense if the tip were scanning *below* the sample instead of *above* the sample, then it would just flip the direction of the electric field caused by the bias, which would be correct. But the `ppafm` code does not allow this anyway, we expect the direction from the tip to the sample to always be a direction of decreasing *z* coordinate. So, under which circumstances could `zpos`<0 happen and why I believe adding `lvec[3,2]` (the lattice vector perpendicular to the slab, essentially) is the right way to correct it then?
Imagine a periodic system with periodicity in *z* given by slab the thickness of, say, `lvec[3,2] = 20.0`. Suppose that the user did not bother to align the position of the sample with the origin of coordinates so the *z* coordinates of the atoms of the sample happen to be somewhere in the interval `10.0 < z < 15.0`. Say `z0 = 14.0` is the reasonable choice to define `zpos` then. Suppose further the user wants to see the forces and LCPD and such in the region starting from 2 angstrom above the sample and going up to 7 angstrom above the sample, which makes if from 17.0 to 22.0 in absolute coordinates. Oops, but the slab ends at `z=20.0`. Does the user have to shift the bottom of the sample from z=10.0 to z=0 and recalculate the Hartree potential and everything in the shifted geometry, so that there would be enough space above the sample? That should not be necessary: Thanks to the periodicity in *z*, the Hartree potential and everything at `z=22.0` should be the same as it is at `z=2.0`. Except that `zpos = z - z0 = 8.0` at `z=22.0` but `zpos = -12` at `z=2.0` and the latter is of course wrong. Adding `lvec[3,2] = 20.0` to `zpos = -12` corrects it to `zpos = 8.0` and everything is fine. Do you see what I mean?
I will be of course adding the lattice vector conditionally, if `zpos` were already positive, nothing would be added.

yakutovicha:
> > Automatic checks failed on code coverage by tests (for the obvious reason that there are indeed no checks for this part of code yet). Can I somehow deactivate the check?
> 
> I was under the impression that we already lifted this restriction.
> 
> @yakutovicha Do you know?

Looks like there is still a problem. I created an issue for that https://github.com/Probe-Particle/ppafm/issues/241. I will try to fix it before our next meeting.

At the moment, just ignore it.

PR #248, opened 2023-12-21
Paper figure

### LINKED ISSUE #246: Add example that plots the main figure of the new paper

NikoOinonen:
Closes #246 

Adds a new example that reproduces the figure in the upcoming paper (#68).

Also adds a new module `ppafm.data` that contains some tools for managing downloading datasets. Should be easy to expand with more datasets in the future. (This is basically copy-pasted from my other project: https://github.com/SINGROUP/ml-spm/blob/main/mlspm/datasets.py)

### DIFF
--- a/doc/sphinx/source/ppafm.rst
+++ b/doc/sphinx/source/ppafm.rst
@@ -84,6 +84,14 @@ ppafm.cpp\_utils module
    :undoc-members:
    :show-inheritance:
 
+ppafm.data module
+---------------------------------
+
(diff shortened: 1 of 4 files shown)


NikoOinonen:
Also, the dataset is up on Zenodo here: https://zenodo.org/records/10418629. It has a DOI (https://doi.org/10.5281/zenodo.10418629), so we could cite it in the paper if we want to.

yakutovicha (approved):
Looks good to me!

PR #251, opened 2024-01-09
Correct sign of densities and potentials in KPFM

### LINKED ISSUE #250: KPFM is wrong because of sign issues
The sign change of the electrostatic potential and charge density, implemented in PR #124, was not followed in the KPFM part of the code. As a result, sign convention in the KPFM part was inconsistent (some potentials being expressed in V, inherited from the main part of the code, others remained in eV; likewise for densities that could be expressed as either charge or electron densities, making a sign difference). KPFM was giving nonsensical results because of this. 

Details shall be obvious from the code changes in a Pull Request I am about to create to address this issue. This issue needs to be settled before it makes any sense to go on with #243 (or even #244, #245).

mondracek:
Fixes #250. What I have done to address the sign issue itself can be best understood from the code changes I think. Besides that

1. The `main` of `ppafm/cli/generateElFF.py` was creating two copies of the electrostatic potential before calculating LCPD: ```v_v0_aux = electrostatic_potential.copy()``` and ```v_v0_aux2 = electrostatic_potential.copy()```. Having the same field written in three places made the code look rather messy to me and I found it difficult to keep track of the required sign changes in such a mess so I did away with those copies. As I understood it, they were needed because the `potential2forces_mem` function in `ppafm/fieldFFT.py` would remove the array with the potential after the potential had been used, unless the function had been called with the `deleteV=False` argument. In order to address this, I have therefore introduced the `deleteV` argument to the function `computeElFF` from `ppafm/HighLevel.py` so that the `deleteV=False` option could be passed from `generateElFF.py` to `potential2forces_mem` through `computeElFF`.
2. The `FFkpfm_tVs0` term was scaled by the tip charge `abs(PPU.params.["charge"])` here https://github.com/Probe-Particle/ppafm/blob/18a5d8a68158e5ffc76d7988bb3e9a6a3b86ff2d/ppafm/HighLevel.py#L160 I did not think this made any sense so I removed this scaling: https://github.com/Probe-Particle/ppafm/blob/aaeedbcbabab2c82453ff283379b4a4e21720071/ppafm/HighLevel.py#L160 @aureliojgc, can you comment on what idea was behind it? Anyway, as I removed it, I changed the meaning of the numerical tip polarizabilities defined here: https://github.com/Probe-Particle/ppafm/blob/aaeedbcbabab2c82453ff283379b4a4e21720071/ppafm/cli/generateElFF.py#L126-L137 I did not change the numbers (only the signs) as it seemed to me that before I removed the scaling, the `FFkpfm_tVs0` contribution was always negligible with respect to the `FFkpfm_t0sV` contribution for default tip polarizability and typical tip charge, which did not seem right. Now this relative contributions of  `FFkpfm_tVs0` and `FFkpfm_tVs0` somewhat improved as to become more equalized.

### DIFF
--- a/ppafm/PPPlot.py
+++ b/ppafm/PPPlot.py
@@ -95,7 +95,7 @@ def plotImages(
         plt.imshow(F[i], origin="lower", interpolation=interpolation, cmap=cmap, extent=extent, vmin=vmin, vmax=vmax)
 
         if cbar:
-            plt.colorbar(shrink=0.5)
+            plt.colorbar(shrink=float(F[i].shape[0]) / float(F[i].shape[1]))
         plotGeom(atoms, bonds, atomSize=atomSize)
         plt.xlabel(r" Tip_x $\AA$")
         plt.ylabel(r" Tip_y $\AA$")
--- a/ppafm/PPPlot.py
+++ b/ppafm/PPPlot.py
@@ -95,7 +95,7 @@ def plotImages(
         plt.imshow(F[i], origin="lower", interpolation=interpolation, cmap=cmap, extent=extent, vmin=vmin, vmax=vmax)
 
         if cbar:
-            plt.colorbar(shrink=float(F[i].shape[0]) / float(F[i].shape[1]))
+            plt.colorbar(shrink=min(1.0, F[i].shape[0] / F[i].shape[1]))
         plotGeom(atoms, bonds, atomSize=atomSize)
         plt.xlabel(r" Tip_x $\AA$")
         plt.ylabel(r" Tip_y $\AA$")
--- a/ppafm/PPPlot.py
+++ b/ppafm/PPPlot.py
@@ -95,7 +95,7 @@ def plotImages(
         plt.imshow(F[i], origin="lower", interpolation=interpolation, cmap=cmap, extent=extent, vmin=vmin, vmax=vmax)
 
         if cbar:
-            plt.colorbar()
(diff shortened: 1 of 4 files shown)


mondracek:
This PR now also closes #252. On the one hand, I understand this is a somewhat non-standard work flow as the two issues, #250 and #252, have nothing in common except they both need to be solved in order to generate the LCPD image for the paper. On the other hand, we need the figure for the paper urgently and the commit that solves #252 changes *literary* just one line of code. As long as the change is uncontroversial, doing all the motions with a separate pull request would be a waste of time and effort.

Tl;dr: Please also check the second commit in this PR.

ProkopHapala (approved):

ppafm/PPPlot.py
```diff
@@ -95,7 +95,7 @@ def plotImages(
         plt.imshow(F[i], origin="lower", interpolation=interpolation, cmap=cmap, extent=extent, vmin=vmin, vmax=vmax)
 
         if cbar:
-            plt.colorbar()
+            plt.colorbar(shrink=float(F[i].shape[0]) / float(F[i].shape[1]) if F[i].shape[0] < F[i].shape[1] else 1.0)
```
NikoOinonen, line 98: I think this would be simpler (and maybe clearer) as:
```suggestion
            plt.colorbar(shrink=min(1.0, F[i].shape[0] / F[i].shape[1]))
```
No need for the `float()` in Python 3.

PR #254, opened 2024-01-22
Add labels to color bars

### LINKED ISSUE #253: Labels need to be added to the color bar in PNG images
The color bar in the PNG figures created by `ppafm-plot-results`, like `df_*.png`, `LCPD _*.png` is labeled by numbers only, without the name of the unit or quantity indicated. I am about to file a PR that will add a "df [Hz]" label for AFM images and "V_LCPD [V]" for LCPD images. More such labels can be added e.g. into tip displacement maps, Laplace-filtered df images, and IETS images, but I would prefer someone else doing it, somebody who actually knows what the units should be in those cases.

mondracek:
Resolves #253

### DIFF
--- a/ppafm/PPPlot.py
+++ b/ppafm/PPPlot.py
@@ -83,6 +83,7 @@ def plotImages(
     atomSize=default_atom_size,
     symmetric_map=False,
     V0=0.0,
+    cbar_label=None,
 ):
     for ii, i in enumerate(slices):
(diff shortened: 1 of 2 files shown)


ondrejkrejci (approved):

ondrejkrejci:
@mondracek - is this still open? Because of IETS and KPFM?

PR #275, opened 2024-04-09
Implement `PpafmParameters` class based on `pydantic.BaseModel`

### LINKED ISSUE #232: Create a class to manage simulation inputs
Currently, all the input parameters are stored in the `params` variable of the `common.py` module. The problem with this approach is that there is only one instance of the module, so one cannot keep running multiple simulations as they might require different inputs.

So, I suggest migrating input configuration into a class and handling things there.

Ideally, after this is implemented the test `tests/_test_vdw.py` can be re-enabled and everything should pass.

- [ ] Move the default `params.ini` file from the `ppafm/defaults/` folder to the wiki to give people an idea of what it looks like.

yakutovicha:
fixes #232 

- [ ] When the PR is about to be merged, update [the defaults](https://github.com/Probe-Particle/ppafm/wiki/Setting-simulation-parameters).
- [x] Add tests for the `PpafmParameters` dataclass.


ondrejkrejci:
@yakutovicha - after the implementation - the note on the `default/params.ini` on this [wiki](https://github.com/Probe-Particle/ppafm/wiki/Setting-simulation-parameters)

yakutovicha:
@NikoOinonen, I guess I can't avoid modifying `AFMulator` as it depends on the functions that were migrated already. Should I proceed?

NikoOinonen:
@yakutovicha Sure, go ahead if it is necessary. Once you are done, I will take a look and maybe make a follow up PR if something should be changed.

yakutovicha:
I believe this is done. Since the change is quite intrusive, I would like several people to take a look at this PR. Please try to break it and share an example of a simulation that was working before and doesn't work now.

ppafm/common.py
```diff
@@ -17,56 +19,61 @@
...
+    charge: float = 0.00
+    Apauli: float = 18.0
+    Bpauli: float = 1.0
+    ffModel: str = "LJ"
+    Rcore: float = 0.7
+    r0Probe: typing.List[float] = [0.00, 0.00, 4.00]
+    stiffness: typing.List[float] = [-1.0, -1.0, -1.0]
+    klat: float = 0.5
+    krad: float = 20.00
+    tip: str = "s"
```
NikoOinonen, line 50: `tip` can also sometimes be an `np.ndarray`. This is for example the case in `generateElFF.py` when the tip density is passed as a `.xsf` file. Can this be a `typing.Union[str, np.ndarray]`?

ppafm/common.py
```diff
@@ -17,56 +19,61 @@
...
-    "#": None,
-}
-# fmt: on
+class PpafmParameters(pydantic.BaseModel):
+    """Parameters for the PPAFM simulation.
+
+    The parameters description can be found on the wiki page: https://github.com/Probe-Particle/ppafm/wiki/Setting-simulation-parameters
+    """
+
+    PBC: bool = True
```
NikoOinonen, line 28: Should we use snake_case for all of these? Now it's a mixture of upper, lower, snake, and CamelCase.

ppafm/common.py
```diff
@@ -104,7 +111,7 @@ class CLIParser(ArgumentParser):
         "Rcore": {
             "action": "store",
             "type": float,
-            "default": params["Rcore"],
+            "default": 0.7,
```
NikoOinonen, line 114: Shouldn't this be `PpafmParameters.Rcore` so that it always uses the same default as above even if it's changed?

ppafm/common.py
```diff
@@ -329,9 +336,9 @@ def get_simple_df_weight(n=10, dz=0.1):
...
 def Fz2df(
     F,
-    dz=0.1,
-    k0=params["kCantilever"],
-    f0=params["f0Cantilever"],
+    dz,
+    k0,
+    f0,
     amplitude=1.0,
     units=16.0217656,
```
NikoOinonen, line 343: This probably does not need to be divided into multiple lines. The comma on the final line is forcing the formatter to keep it on multiple lines.

ppafm/common.py
```diff
@@ -436,90 +443,86 @@ def maxAlongDirEntropy(atoms, hdir, beta=1.0):
...
+def autoGridN(parameters):
+    parameters.gridN[0] = round(np.linalg.norm(parameters.gridA) * 10)
+    parameters.gridN[1] = round(np.linalg.norm(parameters.gridB) * 10)
+    parameters.gridN[2] = round(np.linalg.norm(parameters.gridC) * 10)
+    return parameters.gridN
 
 
 # overide default parameters by parameters read from a file
-def loadParams(fname):
+def loadParams(fname, parameters):
```
NikoOinonen, line 454: Would this function be a better fit as a method of the `PpafmParameters` class? It seems like we always use this function right after constructing the parameters object, so it could even be made into a class method like:
```python
class PpafmParameters(pydantic.BaseModel):
    ...
    @classmethod
    def from_file(cls, file_path="./params.ini"):
        parameters = cls()
        # populate parameters....
        return parameters

parameters = PpafmParameters.from_file("./params.ini")
```

ppafm/common.py
```diff
@@ -436,90 +443,86 @@ def maxAlongDirEntropy(atoms, hdir, beta=1.0):
...
+    parameters.tip = parameters.tip.replace("'", "")
+    # Necessary for working even with quotemarks in params.ini
+    parameters.tip_base[0] = parameters.tip_base[0].replace('"', "")
+    parameters.tip_base[0] = parameters.tip_base[0].replace("'", "")
+    if verbose > 0:
+        print("Loaded parameters from ", fname)
 
 
-def apply_options(opt):
+def apply_options(opt, parameters):
```
NikoOinonen, line 513: Maybe this function as a method of `PpafmParameters` as well?

NikoOinonen (commented):
I didn't test properly yet, but I looked through the changes. I did not see anything wrong, but I left a few comments on things that maybe could be changed.

ppafm/common.py
yakutovicha: done in https://github.com/Probe-Particle/ppafm/pull/275/commits/fe8eb2200d13462bcb1a401dc7087e1a1e83fd7b and in https://github.com/Probe-Particle/ppafm/pull/275/commits/42397888c04666910d8e2571e08c1c9204f4ffb2

ppafm/common.py
yakutovicha: that would be too much for this PR, I would leave this point for a separate issue, if possible.

ppafm/common.py
yakutovicha: done in https://github.com/Probe-Particle/ppafm/pull/275/commits/2e825103ecb25625fe6cd95fa3e614ccf274c855

ppafm/common.py
yakutovicha: done in https://github.com/Probe-Particle/ppafm/pull/275/commits/2e825103ecb25625fe6cd95fa3e614ccf274c855

ppafm/common.py
yakutovicha: done in https://github.com/Probe-Particle/ppafm/pull/275/commits/1f1e6f8f92e40edaf76fe3d3bd278f1dfaf393d3

ppafm/common.py
yakutovicha: done in https://github.com/Probe-Particle/ppafm/pull/275/commits/1903af182ca1ffe78fffe8e3d7fd3087aca5c6aa

ppafm/common.py
```diff
@@ -17,56 +22,155 @@
...
+        suffix = file_path.suffix
+        try:
+            with open(file_path) as file:
+                if suffix == ".ini":
+                    lines = file.readlines()
+                    object.load_ini(lines)
+                elif suffix == ".yaml":
+                    object.apply_options(yaml.safe_load(file))
+                elif suffix == ".toml":
+                    object.apply_options(toml.load(file))
```
NikoOinonen, line 95: Do we need to support both yaml and toml? Having too many options for doing the same thing might just create confusion. I would probably just pick one.

NikoOinonen (changes requested):

ppafm/common.py
```diff
@@ -17,56 +22,155 @@
...
+            raise ValueError(f"File {file_path} not found")
+        return object
+
+    def apply_options(self, obj):
+        for key, value in obj.items():
+            print("Applying", key, value)
+            if hasattr(self, key) and value is not None:
+                setattr(self, key, value)
+
+    def dump_parameters(self, file_path: typing.Union[str, Path] = Path("params.ini")):
```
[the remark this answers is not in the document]
NikoOinonen: Also, we should add a test for these methods. Easiest would be to just create some random parameter object, dump, then load it, and check that the loaded object is identical to the original one.

ppafm/common.py
NikoOinonen: Makes sense.

ppafm/common.py
yakutovicha: We do not, it was just so easy to implement all of them.

I would pick `toml` if we want to support only one. It is very close to the original "ini" format. Also, starting from Python 3.11 is a part of the official Python distribution, so there is no need for an extra module.

Maybe let's hear @ProkopHapala 's opinion here?

ppafm/common.py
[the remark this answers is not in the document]
yakutovicha: For tests - yes. For getting rid of the "ini" format - it would be nice to have some overlap period when both are supported. Also, this functionality allows to perform the conversion. So users can gradually convert their inputs

ppafm/common.py
ProkopHapala: I have no experience with either, but being part of standard python distribution is strong argument for me. My big concern is to minimize external dependencies because I often angry when I install something ending up in dependency hell.

yakutovicha: It looks like the majority leans towards only supporting `toml` format. Then it is decided!

yakutovicha: done in https://github.com/Probe-Particle/ppafm/pull/275/commits/2ba6441df8253fba6b0d931734e0a74f6f311ab9

ppafm/common.py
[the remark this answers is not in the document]
yakutovicha: done in https://github.com/Probe-Particle/ppafm/pull/275/commits/2ba6441df8253fba6b0d931734e0a74f6f311ab9

ppafm/common.py
```diff
@@ -67,7 +67,6 @@ class PpafmParameters(pydantic.BaseModel):
     colorscale_kpfm: str = "seismic"
     ddisp: float = 0.05
     aMorse: float = -1.6
-    tip_base: typing.List = ["None", 0.00]
```
yakutovicha: I had to do it, cause Toml was getting crazy trying to pass an inhomogeneous array.

At the same time, the parameter is not used anywhere (`get grep tip_base`) and was supposed to be [removed](https://github.com/Probe-Particle/ppafm/wiki/Setting-simulation-parameters#temporary-or-past-parameters).

yakutovicha:
I did all I was planning to do, so from my side it is ready.

NikoOinonen (approved):
Everything looks good to me.

PR #281, opened 2024-05-22
Make scanner work when scan_dim is passed as a non-tuple array

NikoOinonen:
The scanner in AFMulator would fail in strange ways when the scan dimension was passed as a list or numpy array instead of a tuple. This PR fixes the problem by simply converting to a tuple on initialization.


ProkopHapala (approved):
nice little improvement. Code seems clear. I don't see anything controversial.

PR #290, opened 2024-07-23
Fix reading atom_frac from geometry.in

### LINKED ISSUE #289: Reading geometry.in with fractional coordinates incorrect
The input from an FHI-AIMS geometry.in file is not read correctly if the atomic coordinates are specified as fractional coordinates (using the `atom_frac` keyword); see e.g. the attached `geometry.in.txt` file here:
geometry.in.txt](https://github.com/user-attachments/files/16351689/geometry.in.txt)
The attached file contains a single layer of FeCl2, the elementary surface unit cell being repeated 5x5. So the structure should appear as a homogeneous hexagonal layer in AFM images (simulated using the Lennard-Jones potentials), but it does not.
I already know what the problem is and how to solve it (essentially a wrong order of matrix multiplication inside the `loadGeometryIN` function in `ppafm/io.py`), so I will assign myself and create the PR to resolve the issue.

mondracek:
Fixes #289

### DIFF
--- a/ppafm/io.py
+++ b/ppafm/io.py
@@ -136,7 +136,7 @@ def loadGeometryIN(fname):
                 elif ws[0] == "lattice_vector":
                     lvec.append([float(ws[1]), float(ws[2]), float(ws[3])])
                 elif ws[0] == "atom_frac" and len(lvec) == 3:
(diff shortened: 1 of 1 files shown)


ondrejkrejci (approved):
image
You are right!

ondrejkrejci:
@mondracek - Looks good to me. Feel free to Squach and merge

PR #292, opened 2024-08-05
Fix automated builds for MacOS.

### LINKED ISSUE #288: MacOS github runner deprecated
The workflow for publishing the PyPI package currently uses the `macos-11` runner for making the MacOS build. It looks like this was just recently deprecated: https://github.blog/changelog/2024-05-20-actions-upcoming-changes-to-github-hosted-macos-runners/. The publishing workflow is now just hanging forever.

There are several replacing runners: `macos-12`, `macos-13`, `macos-14`, and `macos-latest`. I don't use Mac so I don't know what the differences between the versions are. @yakutovicha I think you were using ppafm on a Mac. Do you have any opinion on this?

NikoOinonen:
Fixes #288 

Simply update the version of the runner to latest for MacOS.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -57,36 +57,36 @@ jobs:
                   name: release
                   path: dist/*.tar.gz
 
-    # publish:
+    publish:
 
-    #     name: Build and publish on PyPI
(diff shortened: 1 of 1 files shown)


ondrejkrejci (approved):
Based on your comment in #288 it seems completely legit.

Issue #279, opened 2024-05-17
Continuous-integration tests keep failing, blocking any new pull request


mondracek:
I stumbled over this issue when trying to submit PR #278. This is essentially a copy of what I've reported there:
* The tests fail at the second version of python that is going to be tested, no matter which version it happens to be.
* Therefore, if one modifies `ci.yml` to test only a single version of python, the tests will pass.
* The failure occurs at the `Run pytest` stage, the error message (at lines 31-33) is the following
   - 31 ` /home/runner/work/_temp/6ab19913-178e-4b8f-92bf-dae56f3545bf.sh: line 1:  1858 Segmentation fault      (core dumped) PPAFM_RECOMPILE=1 pytest tests examples -v --cov --cov-report json`
   - 32 `examples/PTCDA_single/example_ptcda.py::example_ptcda_single`
   - 33 `Error: Process completed with exit code 139.`
 * The tests fail even for a branch that is an exact copy of the current `main` branch, so it is not an effect of any changes I tried to implement.
 
I don't have much of an idea how to debug and fix these issues with the GitHub 'workflow' functionality, so I would appreciate any help from you guys who know more about it.

NikoOinonen:
Considering that the tests started failing without any change from our side, it looks like something changed in the Github test runner.

> The tests fail at the second version of python that is going to be tested, no matter which version it happens to be.

If I recall @yakutovicha was saying before that running multiple simulations in a single process can cause failures because the state of the parameters persist between simulations. Maybe the test runner changed so that the different python versions somehow share resources or something?

mondracek:
Yeah, @yakutovicha , do you think this is related to the #232 issue?

NikoOinonen:
I did some more testing, and it looks to me like it's actually specific versions of Python that are failing. 3.8, 3.9, and 3.10 seem to be affected. I was also able to replicate the segfault locally.

NikoOinonen:
Not only different versions of Python, but different patch versions. For example, 3.9.18 works, but 3.9.19 does not, so I think this is why it started to fail suddenly.

This also has something to do with the parameter sharing between simulation runs, because the specific test that fails, only fails if run after another test, but not when run individually. This makes it annoying to try to debug.

NikoOinonen:
[...]

PR #299, opened 2024-08-27
End support for Python 3.7

yakutovicha:
Python 3.8 was put at the end of life on 2024-10-07 [1], so we should also stop supporting this version soon.

[1] https://devguide.python.org/versions/


.github/workflows/ci.yml
```diff
@@ -7,7 +7,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                python-version: ['3.7', '3.9', '3.10', '3.11', '3.12']
+                python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
```
NikoOinonen, line 10: The test for 3.8 seems to be failing still (#228). Probably related to #279, so maybe leave it disabled until we figure it out.

yakutovicha:
Thanks, @NikoOinonen, looks like it is ready then 👍

NikoOinonen (approved):
👍

PR #303, opened 2024-09-18
fix comment handling

### LINKED ISSUE #302: comments in params.ini not recognised
Hi all,

I think the handling of possible comments in `ppafm.common.load_ini` isn't working quite right. For example, if I run the PTCDA_single example and comment out the grid specification like this

```ini
probeType       8                       # atom type of ProbeParticle (to choose L-J potential ),e.g. 8 for CO, 24 for Xe
charge         -0.05                    # effective charge of probe particle [e]
klat            0.37                            # [N/m] harmonic spring potential (x,y) components, x,y is bending stiffnes
krad            20.00                           # [N/m] harmonic spring potential R component, R is the particle-tip bond-length stiffnes
r0Probe         0.0 0.0  4.00                   # [Å] equilibirum position of probe particle (x,y,R) components, R is bond length, x,y introduce tip asymmetry
PBC             True                                    # Periodic boundary conditions ? [ True/False ]
#gridA       20.000   0.0000  0.00000    # a-vector of unit cell; recomanded format (x,y,0)
# gridB        0.000  20.0000  0.00000    # b-vector of unit cell; recomanded format (x,y,0)
# gridC        0.000   0.0000 20.00000    # c-vector of unit cell; recomanded format (0,0,z)
scanMin      0.0    0.0    16.5          # start of scanning (x,y,z)
scanMax     20.0   20.0    22.0          # end of scanning (x,y,z)
Amplitude       1.0                                             # [Å] oscilation amplitude for conversion Fz->df
scanTilt      0.1    0.0   0.02
tiltedScan    True
```

I get an error `ValueError: Parameter #gridA is not known`. (Whether there is a space after # doesn't matter.) 

If I'm reading this right, this could be solved, e.g., by adding 
```python
if key[0] == "#":
    continue
```

after line 121 here. At least the example runs with this fix. I can do this quickly if there's no bigger ideas for this.

https://github.com/Probe-Particle/ppafm/blob/7c0aebb7a1a10357a2817af4f24bf93c1c096e82/ppafm/common.py#L117-L125

LauriKurki:
fixes #302 

I added the check if first character on the line is `#`, and removed the old check since that part can't be reached (the params class doesn't have attribute `#`).

### DIFF
--- a/ppafm/common.py
+++ b/ppafm/common.py
@@ -117,10 +117,10 @@ class PpafmParameters(pydantic.BaseModel):
     def load_ini(self, lines):
         for line in lines:
             words = line.split()
+            if words[0][0] == "#":
+                continue
             if len(words) >= 2:
                 key = words[0]
-                if key[0] == "#":
-                    continue
                 if hasattr(self, key):
                     val = getattr(self, key)
                     if verbose > 0:
--- a/ppafm/common.py
+++ b/ppafm/common.py
@@ -116,9 +116,8 @@ class PpafmParameters(pydantic.BaseModel):
 
     def load_ini(self, lines):
         for line in lines:
+            line = line.split("#")[0].strip()
             words = line.split()
-            if words[0][0] == "#":
-                continue
             if len(words) >= 2:
                 key = words[0]
                 if hasattr(self, key):
--- a/ppafm/common.py
+++ b/ppafm/common.py
@@ -119,10 +119,10 @@ class PpafmParameters(pydantic.BaseModel):
             words = line.split()
             if len(words) >= 2:
                 key = words[0]
+                if key[0] == "#":
+                    continue
                 if hasattr(self, key):
                     val = getattr(self, key)
-                    if key[0][0] == "#":
-                        continue
                     if verbose > 0:
                         print(key, " is class ", val.__class__)
                     if isinstance(val, bool):
--- a/ppafm/common.py
+++ b/ppafm/common.py
@@ -124,36 +124,39 @@ class PpafmParameters(pydantic.BaseModel):
                     val = getattr(self, key)
                     if verbose > 0:
                         print(key, " is class ", val.__class__)
-                    if isinstance(val, bool):
-                        word = words[1].strip()
-                        setattr(self, key, word[0] == "T" or word[0] == "t")
-                        if verbose > 0:
-                            print(key, getattr(self, key), ">>", word, "<<")
-                    elif isinstance(val, float):
-                        setattr(self, key, float(words[1]))
-                        if verbose > 0:
-                            print(key, getattr(self, key), words[1])
-                    elif isinstance(val, int):
-                        setattr(self, key, int(words[1]))
-                        if verbose > 0:
-                            print(key, getattr(self, key), words[1])
-                    elif isinstance(val, str):
-                        setattr(self, key, words[1])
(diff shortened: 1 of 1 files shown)


ppafm/common.py
```diff
@@ -119,10 +119,10 @@ def load_ini(self, lines):
             words = line.split()
             if len(words) >= 2:
                 key = words[0]
+                if key[0] == "#":
+                    continue
                 if hasattr(self, key):
                     val = getattr(self, key)
-                    if key[0][0] == "#":
```
ondrejkrejci, line 124: What if we leave both of these options, before line 120, then it would not matter, if we start with `# someparameter` or `#someparameter` and it also would not matter on the amount of words at all.

ondrejkrejci: Anyway this fix will be short-lived as we are wanting to move towards _toml_ .

LauriKurki: Done in https://github.com/Probe-Particle/ppafm/pull/303/commits/4939276d6d89d758e96d5f5f1738e2904e1835c0

LauriKurki: And yes, I figured that this will be moved to toml in the future but I was working on some PPSTM things and this came up...

NikoOinonen: Does this work if the comment character is somewhere in the middle of the line? It looks like a space before `#` would break it at least.

I guess the idea is just to ignore everything after the comment character. I think this in the beginning of the loop would do it:
```python
line = line.split("#")[0].strip()
```

LauriKurki: Agree that ignoring everything after is the effect we want. However, the parser is quite strict and, e.g., for the grid it gets specifically 3 floats after the key

https://github.com/Probe-Particle/ppafm/blob/7c0aebb7a1a10357a2817af4f24bf93c1c096e82/ppafm/common.py#L143-L145

so comment in the middle of the line breaks the parser if there's not enough values.

```sh
  File "/Users/kurkil1/git/ppafm/ppafm/common.py", line 144, in load_ini
    setattr(self, key, [float(words[1]), float(words[2]), float(words[3])])
IndexError: list index out of range
```

LauriKurki: What would be the expected behaviour if the `params.ini` is not valid, e.g. only 2 values for the grid? Probably possibilities are to revert to default parameters or raise this error.

NikoOinonen: I would say that it needs to raise some error, so that you don't accidentally use the wrong parameters without knowing.

Maybe the error message could be better. We could add a `try-except` block around the `if-elif`-chain and if it runs into an error, print something like `Error parsing parameters on line "[insert offending line]"`.

LauriKurki: > Does this work if the comment character is somewhere in the middle of the line? It looks like a space before `#` would break it at least.
> 
> I guess the idea is just to ignore everything after the comment character. I think this in the beginning of the loop would do it:
> 
> ```python
> line = line.split("#")[0].strip()
> ```

done in https://github.com/Probe-Particle/ppafm/pull/303/commits/c400f3020729d67b05833060730be96878e59726

> try-except block...

done in https://github.com/Probe-Particle/ppafm/pull/303/commits/c91bf6d3cef7add59b3355185a4ed7a5bf175dfe

NikoOinonen (approved):
Thanks, looks good now.

PR #305, opened 2024-09-20
Fix API docs

### LINKED ISSUE #304: Sphinx docs failing
The sphinx API docs at https://ppafm.readthedocs.io/en/latest/index.html are not building correctly at the moment. All of the pages are empty.

NikoOinonen:
Fixes #304 

The API docs are currently blank because the new dependencies from #275 were missing in the ReadTheDocs build.

The badge on the front page does not show the failure, however. This is because the failed imports are treated as warnings instead of errors. So I also added `fail_on_warning: true` to the config and fixed all of the remaining warnings so that there is at least some indication that the build is failing if this happens again in the future.

### DIFF
--- a/doc/sphinx/requirements.txt
+++ b/doc/sphinx/requirements.txt
@@ -7,3 +7,4 @@ sphinx==7.2.6
 furo==2023.9.10
 myst-parser
 pydantic
+toml
--- a/doc/sphinx/requirements.txt
+++ b/doc/sphinx/requirements.txt
@@ -6,3 +6,4 @@ ase
 sphinx==7.2.6
 furo==2023.9.10
 myst-parser
+pydantic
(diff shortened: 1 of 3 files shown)


doc/sphinx/requirements.txt
```diff
@@ -6,3 +6,6 @@ ase
 sphinx==7.2.6
 furo==2023.9.10
 myst-parser
+pydantic
+toml
+PyQt5
```
ondrejkrejci, line 11: Is PyQT5 now a part of the full installation automatically?

NikoOinonen: No, this is installed just by the automated system that builds the HTML pages at ReadTheDocs. It's because during the build it imports the `GUIWidgets` module which in turn imports `PyQt5`, which was producing a warning.

ondrejkrejci (approved):
Looks good to me.

PR #309, opened 2024-10-07
Fix spline interpolation segfaults.

### LINKED ISSUE #279: Continuous-integration tests keep failing, blocking any new pull request

### LINKED ISSUE #308: Problems with spline interpolation code.
While debugging the CI tests (#279) I found at least two different issues in the spline interpolation C++ code that is causing segmentation faults.

First, on this line of code, the variable `i` seems to be uninitialized, and sometimes can be some large value that causes an out of bounds access later: https://github.com/Probe-Particle/ppafm/blob/17ee88460f0614843d1740c4d6e0dacbc0a9758f/ppafm/cpp/ProbeParticle.cpp#L211 @ProkopHapala Can you check what this is supposed to be, since you wrote it.

The other issue is that once the spline parameters are set using `ppafm.core.setTipSpline`, the spline interpolation remains in effect for the whole rest of the program, because the code only checks whether the pointer exists when deciding what algorithm to use: https://github.com/Probe-Particle/ppafm/blob/17ee88460f0614843d1740c4d6e0dacbc0a9758f/ppafm/cpp/ProbeParticle.cpp#L283-L287 This leads to two problems:
1. The parameter arrays don't necessarily exist anymore (Python garbage collector?), so the pointer could be pointing to an invalid location. This seems to be happening in the current CI test setup that is randomly giving seg faults because of this.
2. If you run multiple calculations in a row, which is now possible after we rearranged all the things, you could be using the wrong algorithm without knowing, if the spline was set somewhere else. I suggest that we add a corresponding `else` block in `perform_relaxation` that explicitly sets the spline parameters to `None` when they were not specified: https://github.com/Probe-Particle/ppafm/blob/17ee88460f0614843d1740c4d6e0dacbc0a9758f/ppafm/HighLevel.py#L138

Any thoughts @ProkopHapala @ondrejkrejci @yakutovicha @mondracek?

NikoOinonen:
Fixes #308
Fixes #279 

This PR moves the spline parameters in the `TIP` namespace into a struct in the C++ code and introduces a corresponding Python class for managing the parameter arrays. A pointer to the parameters are now explicitly passed into the relevant functions instead of being set on a global level. This gets rid of the problem where the parameter array pointer could still be set even after the arrays were freed.

Also fixes one uninitialized variable in the spline interpolation function that was sometimes segfaulting.

All tests are now enabled and seem to be passing without problems again.

### DIFF
--- a/ppafm/HighLevel.py
+++ b/ppafm/HighLevel.py
@@ -103,7 +103,7 @@ def relaxedScan3D(xTips, yTips, zTips, trj=None, bF3d=False):
     return fzs, PPpos
 
 
-def relaxedScan3D_omp(xTips, yTips, zTips, trj=None, bF3d=False):
+def relaxedScan3D_omp(xTips, yTips, zTips, trj=None, bF3d=False, spline_parameters=None):
     if verbose > 0:
         print(">>BEGIN: relaxedScan3D_omp()")
     if verbose > 0:
@@ -117,7 +117,7 @@ def relaxedScan3D_omp(xTips, yTips, zTips, trj=None, bF3d=False):
     rTips[:, :, :, 0] = xTips[:, None, None]
     rTips[:, :, :, 1] = yTips[None, :, None]
     rTips[:, :, :, 2] = zTips[::-1][None, None, :]
-    core.relaxTipStrokes_omp(rTips, rs, fs)
+    core.relaxTipStrokes_omp(rTips, rs, fs, spline_parameters=spline_parameters)
     # rs[:,:,:,:] = rs[:,:,::-1,:].transpose(2,1,0,3).copy()
     # fs[:,:,:,:] = fs[:,:,::-1,:]
     # fzs = fs[:,:,::-1,2].transpose(2,1,0).copy()
@@ -131,26 +131,22 @@ def relaxedScan3D_omp(xTips, yTips, zTips, trj=None, bF3d=False):
     return fzs, rs
 
 
-def perform_relaxation(lvec, FFLJ, FFel=None, FFpauli=None, FFboltz=None, FFkpfm_t0sV=None, FFkpfm_tVs0=None, tipspline=None, bPPdisp=False, bFFtotDebug=False, parameters=None):
+def perform_relaxation(
+    lvec,
+    FFLJ,
+    FFel=None,
+    FFpauli=None,
+    FFboltz=None,
+    FFkpfm_t0sV=None,
+    FFkpfm_tVs0=None,
+    spline_parameters=None,
+    bPPdisp=False,
+    bFFtotDebug=False,
+    parameters=None,
+):
     global FF  # We need FF global otherwise it is garbage collected and program crashes inside C++ e.g. in stiffnessMatrix()
     if verbose > 0:
         print(">>>BEGIN: perform_relaxation()")
-    if tipspline is not None:
-        try:
-            if verbose > 0:
-                print(" loading tip spline from " + tipspline)
-            S = np.genfromtxt(tipspline)
-            xs = S[:, 0].copy()
-            if verbose > 0:
-                print("xs: ", xs)
-            ydys = S[:, 1:].copy()
-            if verbose > 0:
-                print("ydys: ", ydys)
-            core.setTipSpline(xs, ydys)
-        except:
-            if verbose > 0:
-                print("cannot load tip spline from " + tipspline)
-            sys.exit()
     xTips, yTips, zTips, lvecScan = PPU.prepareScanGrids(parameters=parameters)
     FF = FFLJ.copy()
     if FFel is not None:
@@ -180,7 +176,7 @@ def perform_relaxation(lvec, FFLJ, FFel=None, FFpauli=None, FFboltz=None, FFkpfm
     if parameters.tiltedScan:
         trj = trjByDir(len(zTips), d=parameters.scanTilt, p0=[0.0, 0.0, parameters.scanMin[2] - lvec[0, 2]])
     # fzs, PPpos = relaxedScan3D(xTips - lvec[0, 0], yTips - lvec[0, 1], zTips - lvec[0, 2], trj=trj, bF3d=parameters.tiltedScan)
-    fzs, PPpos = relaxedScan3D_omp(xTips - lvec[0, 0], yTips - lvec[0, 1], zTips - lvec[0, 2], trj=trj, bF3d=parameters.tiltedScan)
+    fzs, PPpos = relaxedScan3D_omp(xTips - lvec[0, 0], yTips - lvec[0, 1], zTips - lvec[0, 2], trj=trj, bF3d=parameters.tiltedScan, spline_parameters=spline_parameters)
 
     # transform probe-particle positions back to the original coordinates
     PPpos[:, :, :, 0] += lvec[0, 0]
--- a/ppafm/HighLevel.py
+++ b/ppafm/HighLevel.py
@@ -103,7 +103,7 @@ def relaxedScan3D(xTips, yTips, zTips, trj=None, bF3d=False):
     return fzs, PPpos
 
 
-def relaxedScan3D_omp(xTips, yTips, zTips, trj=None, bF3d=False, spline_parameters=None):
+def relaxedScan3D_omp(xTips, yTips, zTips, trj=None, bF3d=False, tip_spline=None):
     if verbose > 0:
         print(">>BEGIN: relaxedScan3D_omp()")
     if verbose > 0:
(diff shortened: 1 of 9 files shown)


NikoOinonen:
@ProkopHapala The way I did it is maybe not exactly what you were suggesting, but I think this should work quite nicely.  It's now more explicit when the spline interpolation is used. The downside is of course that there is more boilerplate passing of the parameters in several functions, but maybe it's acceptable. See if you like it or not.

tests/human_eye/test_TipForce.py
```diff
@@ -7,46 +7,53 @@
...
+    4.0   0.0 -0.1
+    5.0  -4.0 -4.0
+    10.0 -24.0 -4.0
+    """
+    # S = np.genfromtxt('TipRSpline.ini')
+    S = np.genfromtxt(StringIO(spline_ini), skip_header=1)
+
+    parameters = common.PpafmParameters()
+
+    print("TipRSpline.ini overrides harmonic tip")
```
ProkopHapala, line 25: hmm, so the user should do this manually? I think it would be nice if the spline-class can initialized itself form `spline_ini` file, just for convenience

tests/human_eye/test_TipForce.py
```diff
@@ -7,46 +7,53 @@
...
+    xs = S[:, 0].copy()
+    print("xs: ", xs)
+    ydys = S[:, 1:].copy()
+    print("ydys: ", ydys)
+    plt.plot(xs, ydys[:, 0], "o")
+    plt.plot(xs, ydys[:, 0] + ydys[:, 1], ".")
+    spline_parameters = core.SplineParameters(xs, ydys)
```
ProkopHapala, line 32: I mean it would be convenient to call `core.SplineParameters(fname='tip_spline.dat')` instead

tests/human_eye/test_vdwDamping.py
ProkopHapala, line 1: this is just some formating? I don't see relation to spline (?)

ProkopHapala (approved):
OK

ppafm/core.py
```diff
@@ -150,14 +150,28 @@ def setTip(lRadial=None, kRadial=None, rPP0=None, kSpring=None, parameters=None)
...
+    _fields_ = [
+        ("rff_n", c_int),
+        ("rff_xs", POINTER(c_double)),
+        ("rff_ydys", POINTER(c_double)),
+    ]
 
+    def __init__(self, xs, ydys):
+        super().__init__()
+        self.rff_n = len(xs)
+        self.rff_xs = np.ctypeslib.as_ctypes(xs.astype(np.double))
```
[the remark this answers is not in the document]
NikoOinonen: I originally tried to do it with the `np.ctypeslib.ndpointer` so that you could just use the numpy array, but for some reason that does not work with structs, so I opted to use the `ctypes` pointer instead which works.

I changed it so that the original numpy arrays are stored there as well in a separate variables starting with `np_`.

ppafm/HighLevel.py
```diff
@@ -131,26 +131,22 @@ def relaxedScan3D_omp(xTips, yTips, zTips, trj=None, bF3d=False):
...
-def perform_relaxation(lvec, FFLJ, FFel=None, FFpauli=None, FFboltz=None, FFkpfm_t0sV=None, FFkpfm_tVs0=None, tipspline=None, bPPdisp=False, bFFtotDebug=False, parameters=None):
+def perform_relaxation(
+    lvec,
+    FFLJ,
+    FFel=None,
+    FFpauli=None,
+    FFboltz=None,
+    FFkpfm_t0sV=None,
+    FFkpfm_tVs0=None,
+    spline_parameters=None,
```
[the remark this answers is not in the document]
NikoOinonen: I changed the name to `tip_spline`.

tests/human_eye/test_TipForce.py
NikoOinonen: No, actually I also made it so that you can load it directly from a file using `core.SplineParameters.from_file`. This was already used in `relaxed_scan.py`. I now changed it in this test script as well.

tests/human_eye/test_TipForce.py
NikoOinonen: Now it's like this.

tests/human_eye/test_vdwDamping.py
NikoOinonen: Yes, it's just moving the code inside a function block, which makes it work better with the automated test system.

NikoOinonen:
> On the other hand I agree that it was introducing some hidden variable which was not obvisous/explicit.

Yes, this is exactly the issue, the hidden nature of it made it difficult to know what the state of the data is at any given moment. Especially since the pointers were set on the C++ side, then it's impossible know from the Python side what pointers exist and don't exist.

Actually it's kind of this way with the force field / potential grids as well. Maybe these could be made into something that can be managed on the Python side as well.

> Usually I try to avoid changes to function interfaces especially on the python/C++ interface, and keep the argument list short.

If we have some common parameter lists that we pass to the functions, then maybe these could be packed into structs as well to keep the function signatures shorter. The structs can be easily extended as well without changing the function signature.

NikoOinonen:
@ProkopHapala, you already approved this, but since I made some changes, I'll wait a bit if you want to comment on something before I merge.

ppafm/core.py
```diff
@@ -150,14 +150,30 @@ def setTip(lRadial=None, kRadial=None, rPP0=None, kSpring=None, parameters=None)
...
+
+    def __init__(self, xs, ydys):
+        super().__init__()
+        self.np_rff_xs = np.ascontiguousarray(xs, dtype=np.double)
+        self.np_rff_ydys = np.ascontiguousarray(ydys, dtype=np.double)
+        self.rff_n = len(self.np_rff_xs)
+        self.rff_xs = np.ctypeslib.as_ctypes(self.np_rff_xs)
+        self.rff_ydys = np.ctypeslib.as_ctypes(self.np_rff_ydys.reshape(-1))
+
+    @classmethod
```
ProkopHapala, line 173: I did not know this pattern `@classmethod` before. Basically it is like overloaded constructor (?)

ProkopHapala (approved):
OK

ProkopHapala:
> u already approved this, but since I made some changes, I'll wait a bit if you want to comment on something before I merge.

I think it is good. 
(In general, I don't want to block you, inc case I forget to look on it.)

ProkopHapala:
> Especially since the pointers were set on the C++ side, then it's impossible know from the Python side what pointers exist and don't exist.

Yes, I think if we use something else than `ctypes` for interfacing (e.g. [pybind11](https://github.com/pybind/pybind11) ? ) this can be better. But I wanted something which does not have dependences, and it is general (even if we cannot the C++ core to e.g. Julia. Again historically I considered python just as some script to make working with the C++ core more convenient)

> Actually it's kind of this way with the force field / potential grids as well. Maybe these could be made into something that can be managed on the Python side as well.

Talking about big structural changes, it would be nice to make everything into classes inside C/C++ so that we can dynamically create and destroy instances of ppafm simulator and keep multiple different instances in memory simultanously. I think for some applications it could be useful.

I that case I would make python classes which mimic the C++ classes and keep the same data fields, and destroy/nullify them in sync.

> If we have some common parameter lists that we pass to the functions, then maybe these could be packed into structs as well to keep the function signatures shorter. The structs can be easily extended as well without changing the function signature.

perhaps yes. At the beggining I did not want to use structs because it looks a bit more complicated and add some boilerplate. But form some complexity it is probably worth it.

ppafm/core.py
NikoOinonen: Kind of, yes. It makes the method so that it can be called without creating an instance of the class, and the first argument, instead of being `self` pointing to the instance, is `cls` which points to the class constructor.

NikoOinonen:
> I that case I would make python classes which mimic the C++ classes and keep the same data fields, and destroy/nullify them in sync.

> I think if we use something else than ctypes for interfacing (e.g. [pybind11](https://github.com/pybind/pybind11) ? )

Yes, if we want to do this, then `ctypes` does not work, since it only interfaces with pure C, not C++ features like classes. `pybind11` is advertising interfacing with C++, so that would probably be the way to go.

Issue #312, opened 2024-10-09
FDBM (Full-density based method) broken by recent changes


mondracek:
The electrostatic and vdW calculations do not currently work in FDBM, as you can test e.g. with the `ppafm/examples/pyridineDensOverlap/` example.

Running `ppafm-generate-elff -i sample/LOCPOT.xsf --tip_dens tip/CHGCAR.xsf --Rcore 0.7 --doDensity` leads to the error message
``NameError: name 'parameters' is not defined`` 
issued inside the `ppafm/ppafm/HighLevel.py/getAtomsWhichTouchPBCcell` function when trying to call
``atoms, nDim, lvec = io.loadGeometry(fname, format=geometry_format, parameters=parameters)``

Similarly, calling `ppafm-generate-dftd3 -i sample/LOCPOT.xsf --df_name PBE` gives
``ValueError: Please provide the parameters dictionary here``,
raised inside the `loadGeometry` function itself, after this function has been called from `ppafm/HighLevel.py/computeDFTD3` by the line
``atoms, nDim, lvec = io.loadGeometry(input_file, format=geometry_format, parameters=parameters)``

NikoOinonen:
We probably should add the pyridine example into the CI tests to make sure this does not happen again.

yakutovicha:
> We probably should add the pyridine example into the CI tests to make sure this does not happen again.

Thanks, @NikoOinonen, I assume you are referring to [this one](https://github.com/Probe-Particle/ppafm/tree/main/examples/pyridineDensOverlap)?

NikoOinonen:
> Thanks, @NikoOinonen, I assume you are referring to [this one](https://github.com/Probe-Particle/ppafm/tree/main/examples/pyridineDensOverlap)?

@yakutovicha Yes, exactly.

yakutovicha:
> > Thanks, @NikoOinonen, I assume you are referring to [this one](https://github.com/Probe-Particle/ppafm/tree/main/examples/pyridineDensOverlap)?
> 
> @yakutovicha Yes, exactly.

done in https://github.com/Probe-Particle/ppafm/pull/316/commits/e45868a281c4f00f6fad949fedff2c02a5c600d4

PR #313, opened 2024-10-18
Add density cutoff option to fix FDBM for all-electron densities.

### LINKED ISSUE #276: Fixing FHI-aims electron densities for FDBM
I have been having some problems using electron densities from FHI-aims with the full-density-based model. The problem shows up in the simulated images as severe noise and artifacts in the mid to far range images.

Here is a random test molecule from our database which exhibits the issue:
image
Far:

Close:

The close range image is fine, but in the far-range image there is a lot of noise and a weird round artifact at the position of the Cl atom.

Since the problem only shows up at the far range, I first thought the issue might be something to do with the integration grid in aims, which gets more sparse further away from the atoms. However, after a quite a bit of testing, it looks like the issue actually stems from the core electrons. FHI-aims is an all-electron code so the density has sharp spikes at the positions of the atoms. 

By simply applying a cutoff to the electron density, the issue seems to go away:

After trying a bunch of different cutoff values, it looks like somewhere around 100 the image becomes completely smooth for this particular case. Interestingly, going down even far as 1 with the cutoff still does not affect the close range image in any appreciable way. For reference, the original value at the position of the Cl atom is ~18000.

A line scan through the center of the Cl atoms:
density_line

It seems worth implementing this cutoff into ppafm in some way in order to make electron densities from aims also usable with the FDBM.

@ProkopHapala Do you foresee any problem with applying a cutoff in this way? It does get rid of some electrons, but then again, previously we have used the pseudo-potential densities that don't have the core electrons either.

NikoOinonen:
Fixes #276 

- [x] Add an example of a full-density-based model using AIMS electron density.
- [x] Test the example with CLI once #316 gets fixed.
- [x] Update dropbox links to Zenodo once #314 is fixed.

Adds an option to cut off high-density values in FDBM Pauli integral. This enables using the FDBM simulation with all-electron densities where high-density values near nuclei cause artifacts in the resulting images.

### DIFF
--- a/ppafm/cli/gui/ppafm_gui.py
+++ b/ppafm/cli/gui/ppafm_gui.py
@@ -74,6 +74,7 @@ TTips = {
     "Rotation": "Rotation: Set sample counter-clockwise rotation angle around center of atom coordinates.",
     "fdbm_V0": "FDBM V0: Prefactor in Pauli interaction integral in the full-density based model.",
     "fdbm_alpha": "FDBM alpha: Exponent in Pauli interaction integral in the full-density based model.",
+    "fdbm_cutoff": "Density cutoff: cut out high values of the electron density. Useful when working with all-electron densities where extremely high values can arise near nuclei, causing artifacts in the images",
     "PBCz": "z periodicity: When checked, the lattice is also periodic in z direction. This is usually not required, since the scan is aligned with the xy direction.",
     "PBC": "Periodic Boundaries: Lattice vectors for periodic images of atoms. Does not affect electrostatics calculated from a Hartree potential file, which is always assumed to be periodic.",
     "k": "k: Cantilever spring constant. Only appears as a scaling constant.",
@@ -123,11 +124,13 @@ class ApplicationWindow(QtWidgets.QMainWindow):
         self.Zs = None
         self.qs = None
         self.rho_sample = None
+        self.rho_sample_original = None  # Stores the original density when a cutoff is applied
         self.rho_tip = None
         self.rho_tip_delta = None
         self.sample_lvec = None
         self.rot = np.eye(3)
         self.df_points = []
+        self.previous_cutoff = None
 
         # Initialize OpenCL environment on chosen device and create an afmulator instance to use for simulations
         oclu.init_env(device)
@@ -302,10 +305,33 @@ class ApplicationWindow(QtWidgets.QMainWindow):
         if self.verbose > 0:
             print("updateParams", Q, sigma, multipole, tipStiffness, tipR0, use_point_charge, A_pauli, B_pauli, type(self.qs))
 
-        self.afmulator.iZPP = int(self.bxZPP.value())
         if self.rho_sample is not None:  # Use FDBM
-            if self.afmulator.B_pauli != B_pauli:
-                self.afmulator.setRho(self.rho_tip, B_pauli=B_pauli)
+
+            if self.bxCutoffEnable.isChecked():
+                self.bxCutoffValue.setDisabled(False)
+                cutoff = float(self.bxCutoffValue.value())
+            else:
+                self.bxCutoffValue.setDisabled(True)
+                cutoff = None
+
+            if cutoff is None or self.previous_cutoff is None:
+                cutoff_changed = cutoff != self.previous_cutoff
+            else:
+                cutoff_changed = not np.isclose(cutoff, self.previous_cutoff)
+            self.previous_cutoff = cutoff
+
+            if cutoff_changed:
+                if cutoff is not None:
+                    self.rho_sample = self.rho_sample_original.clamp(maximum=cutoff, in_place=False)
+                else:
+                    self.rho_sample = self.rho_sample_original
+
+            if self.afmulator.B_pauli != B_pauli or cutoff_changed:
+                if cutoff is not None:
+                    rho_tip = self.rho_tip.clamp(maximum=cutoff, in_place=False)
+                else:
+                    rho_tip = self.rho_tip
+                self.afmulator.setRho(rho_tip, B_pauli=B_pauli)
         else:
             if self.rho_tip is None:  # else we just keep the tip density loaded from file
                 if use_point_charge:
@@ -313,6 +339,8 @@ class ApplicationWindow(QtWidgets.QMainWindow):
                     self.afmulator.setRho(None, sigma)
                 elif isinstance(self.qs, HartreePotential):
                     self.afmulator.setRho({multipole: Q}, sigma)
+
+        self.afmulator.iZPP = int(self.bxZPP.value())
         self.afmulator.scanner.stiffness = np.array(tipStiffness, dtype=np.float32) / -PPU.eVA_Nm
         self.afmulator.tipR0 = tipR0
         self.afmulator.A_pauli = A_pauli
@@ -471,6 +499,8 @@ class ApplicationWindow(QtWidgets.QMainWindow):
             self.qs.release()
         if self.rho_sample is not None:
             self.rho_sample.release()
+        if self.rho_sample_original is not None:
+            self.rho_sample_original.release()
         if self.rho_tip is not None:
             self.rho_tip.release()
         if self.rho_tip_delta is not None:
@@ -502,6 +532,7 @@ class ApplicationWindow(QtWidgets.QMainWindow):
         # Load auxiliary files (if any)
         if rho_sample:
             self.rho_sample, _, _ = ElectronDensity.from_file(rho_sample)
+            self.rho_sample_original = self.rho_sample
         else:
             self.rho_sample = None
         if rho_tip:
@@ -577,9 +608,14 @@ class ApplicationWindow(QtWidgets.QMainWindow):
         if self.rho_sample is None:
             self.bxV0.setDisabled(True)
             self.bxAlpha.setDisabled(True)
+            self.bxCutoffEnable.setDisabled(True)
+            self.bxCutoffEnable.setChecked(False)
+            self.bxCutoffValue.setDisabled(True)
         else:
             self.bxV0.setDisabled(False)
             self.bxAlpha.setDisabled(False)
+            self.bxCutoffEnable.setDisabled(False)
+            self.previous_cutoff = None
 
         # Create geometry editor widget
         self.createGeomEditor()
@@ -999,6 +1035,18 @@ class ApplicationWindow(QtWidgets.QMainWindow):
         lb.setToolTip(TTips["fdbm_alpha"])
         l.addWidget(lb)
         self.bxAlpha = _spin_box((0.1, 9.9), 1.0, 0.02, self.updateParams, TTips["fdbm_alpha"], l)
+
+        vb = QtWidgets.QHBoxLayout()
+        self.l0.addLayout(vb)
+        self.bxCutoffEnable = QtWidgets.QCheckBox()
+        self.bxCutoffEnable.setChecked(False)
+        self.bxCutoffEnable.toggled.connect(self.updateParams)
+        vb.addWidget(self.bxCutoffEnable)
+        lb = QtWidgets.QLabel("Density cutoff")
+        lb.setToolTip(TTips["fdbm_cutoff"])
+        vb.addWidget(lb)
+        self.bxCutoffValue = _spin_box((0.1, 10000.0), 100.0, 10.0, self.updateParams, TTips["fdbm_cutoff"], vb)
+        self.bxCutoffValue.setDisabled(True)
 
     def _create_scan_settings_ui(self):
         # Title
--- a/ppafm/cli/gui/ppafm_gui.py
+++ b/ppafm/cli/gui/ppafm_gui.py
@@ -74,7 +74,7 @@ TTips = {
     "Rotation": "Rotation: Set sample counter-clockwise rotation angle around center of atom coordinates.",
     "fdbm_V0": "FDBM V0: Prefactor in Pauli interaction integral in the full-density based model.",
     "fdbm_alpha": "FDBM alpha: Exponent in Pauli interaction integral in the full-density based model.",
-    "fdbm_cutoff": "Density cutoff: cut out high values of the electron density. Useful when working with all-electron densities where extremely high values can arise near nuclei, causing artifacts in the images",
+    "fdbm_cutoff": "Density cutoff: cut out high values of the electron density. Useful when working with all-electron densities where extremely high values can arise near nuclei, causing artifacts in the images.",
     "PBCz": "z periodicity: When checked, the lattice is also periodic in z direction. This is usually not required, since the scan is aligned with the xy direction.",
     "PBC": "Periodic Boundaries: Lattice vectors for periodic images of atoms. Does not affect electrostatics calculated from a Hartree potential file, which is always assumed to be periodic.",
     "k": "k: Cantilever spring constant. Only appears as a scaling constant.",
--- a/ppafm/cli/gui/ppafm_gui.py
+++ b/ppafm/cli/gui/ppafm_gui.py
@@ -181,7 +181,7 @@ class ApplicationWindow(QtWidgets.QMainWindow):
         self._create_buttons_ui()
 
         # Create curve plotting window
-        self.figCurv = guiw.PlotWindow(parent=self, width=5, height=4, dpi=100)
+        self.figCurv = guiw.PlotWindow(parent=self, width=5, height=4, dpi=100, xlabel="z_tip (Å)", ylabel="df (Hz)")
 
         if input_files:
             self.loadInput(**input_files)
--- a/ppafm/cli/gui/ppafm_gui.py
+++ b/ppafm/cli/gui/ppafm_gui.py
@@ -144,6 +144,7 @@ class ApplicationWindow(QtWidgets.QMainWindow):
         self.afmulator.forcefield.verbose = verbose
         self.afmulator.scanner.verbose = verbose
         FFcl.bRuntime = verbose > 1
+        self.afmulator.bRuntime = verbose > 1
 
         # --- init QtMain
         QtWidgets.QMainWindow.__init__(self)
--- a/ppafm/ml/Generator.py
+++ b/ppafm/ml/Generator.py
@@ -342,6 +342,8 @@ class GeneratorAFMtrainer:
             is FDBM, where it is used for calculating the electrostatic interaction.
         ignore_elements: list of int. Atomic numbers of elements to ignore in scan window distance and position calculations.
             Useful for example for ignoring surface slab atoms in centering the scan window.
+        density_cutoff: float or None. If not None, apply a cutoff to electron densities in the FDBM Pauli integral. Ignored when
+            sim_type is not ``'FDBM'``.
     """
 
     bRuntime = False
@@ -361,6 +363,7 @@ class GeneratorAFMtrainer:
         rhos=None,
         rho_deltas=None,
         ignore_elements=[],
+        density_cutoff=None,
     ):
         self.afmulator = afmulator
         self.aux_maps = aux_maps
@@ -370,8 +373,10 @@ class GeneratorAFMtrainer:
         self.distAbove = distAbove
         self.distAboveActive = distAbove
         self.iZPPs = iZPPs
-        self._prepareBuffers(rhos, rho_deltas)
         self.ignore_elements = ignore_elements
+
+        self.density_cutoff = density_cutoff if self.sim_type == "fdbm" else None
+        self._prepare_tip_buffers(rhos, rho_deltas, self.density_cutoff)
 
         if Qs is None or QZs is None:
             if self.sim_type == "lj+pc":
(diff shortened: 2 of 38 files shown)


ppafm/ml/Generator.py
```diff
@@ -389,29 +395,33 @@ def __init__(
         self.df_steps = self.afmulator.df_steps
         self.z_size = self.scan_dim[2] - self.df_steps + 1
 
-    def _prepareBuffers(self, rhos=None, rho_deltas=None):
+    def _prepare_tip_buffers(self, rhos=None, rho_deltas=None, density_cutoff=None):
```
ondrejkrejci, line 398: Jush checking - is the KPFM procedure using buffers as well @mondracek? Do you have an idea about it? Because if so, then restriction to Tip is maybe not what you want?

ppafm/ml/Generator.py
```diff
@@ -389,29 +395,33 @@ def __init__(
...
         if rhos is None:
-            self.rhos = self.ffts = [(None, None)] * len(self.iZPPs)
+            self.rhos_tip = self.rhos_tip_delta = self.ffts_tip = self.ffts_tip_delta = [None] * len(self.iZPPs)
```
ondrejkrejci, line 400: There is a big difference now, from `[(None, None)] * len(self.iZPPs)` to `[None] * len(self.iZPPs)` Are you exchanging an original double list-like objects with 2 single-list object?.
I am also afraid by the connection of many variables together.

ppafm/ml/Generator.py
```diff
@@ -389,29 +395,33 @@ def __init__(
...
-                rhos_.append(self.afmulator.forcefield.rho_delta)
-                ffts_.append(self.afmulator.forcefield.fft_corr_delta)
-            else:
-                rhos_.append(None)
-                ffts_.append(None)
-            self.rhos.append(rhos_)
-            self.ffts.append(ffts_)
+        self.rhos_tip = []
+        self.ffts_tip = []
+        self._rhos_original = []
```
ondrejkrejci, line 408: Why `_rhos_original` is necessary?

ppafm/ml/Generator.py
```diff
@@ -615,6 +630,26 @@ def handle_distance(self):
             (sw[1][0], sw[1][1], z_min + self.scan_size[2]),
         )
 
+    def set_fdbm_parameters(self, A_pauli, B_pauli):
```
ondrejkrejci, line 633: Where this was set originally? I cannot see it anywhere.

ppafm/ocl/AFMulator.py
```diff
@@ -269,9 +274,10 @@ def setRho(self, rho=None, sigma=0.71, B_pauli=1.0):
                 print("AFMulator.setRho: Preparing buffers")
             if not np.allclose(B_pauli, 1.0):
                 rho_power = self.rho.power_positive(p=self.B_pauli, in_place=False)
-                self.rho.release()  # Let's not keep the original array in device memory to minimize memory foot print
+                if self.minimize_memory:
```
ondrejkrejci, line 277: I am afraid that this way, one loses the original FF, when changes the cutoff parameters.

ppafm/ocl/field.py
```diff
@@ -277,6 +277,70 @@ def to_file(self, file_path, clamp=None):
...
+            # fmt: off
+            cl_program.clamp_hard(queue, global_size, local_size,
+                array_in,
+                grid_out.cl_array,
+                n,
+                minimum,
+                maximum,
+            )
+            # fmt: on
+        elif clamp_type == "soft":
```
ondrejkrejci, line 328: nice possibility of the soft clamping - is it useable through CLI?

ppafm/ocl/field.py
```diff
@@ -292,25 +356,21 @@ def add_mult(self, array, scale=1.0, in_place=True, local_size=(32,), queue=None
...
-        queue = queue or oclu.queue
-        if in_place:
-            grid_out = self
-            array_out = array_in1
-            self._array = None  # The current numpy array will be wrong after operation so reset it
-        else:
-            array_out = cl.Buffer(self.ctx, cl.mem_flags.READ_WRITE, size=array_in1.size)
-            array_type = type(self)  # This way so inherited classes return their own class type
-            grid_out = array_type(array_out, lvec=self.lvec, shape=self.shape, ctx=self.ctx)
+        grid_out = self._prepare_same_size_output_grid(array_in1, in_place)
```
ondrejkrejci, line 362: Are you sure, that removing the queue from here and adding it after preparation of grid out is safe?

ppafm/ocl/field.py
```diff
@@ -277,6 +277,70 @@ def to_file(self, file_path, clamp=None):
             io.save_vec_field(file_head, array[:, :, :, :3], self.lvec, data_format=ext)
             io.save_scal_field(file_head + "_w", array[:, :, :, 3], self.lvec, data_format=ext)
 
+    def _prepare_same_size_output_grid(self, array_in, in_place):
+        if in_place:
+            grid_out = self
```
ondrejkrejci, line 282: array_out = array_in ? Isn't it missing.

ppafm/ocl/field.py
```diff
@@ -292,25 +356,21 @@ def add_mult(self, array, scale=1.0, in_place=True, local_size=(32,), queue=None
...
+
+        queue = queue or oclu.queue
         global_size = [int(np.ceil(n / local_size[0]) * local_size[0])]
+
         # fmt: off
         cl_program.addMult(queue, global_size, local_size,
             array_in1,
             array_in2,
-            array_out,
+            grid_out.cl_array,
```
ondrejkrejci, line 373: Why this change?

ppafm/ocl/field.py
```diff
@@ -357,12 +409,13 @@ def power_positive(self, p=1.2, normalize=True, in_place=True, local_size=(32,),
...
 
-        n = np.int32(array_in.size / 4)
+        queue = queue or oclu.queue
         global_size = [int(np.ceil(n / local_size[0]) * local_size[0])]
+
         # fmt: off
         cl_program.power(queue, global_size, local_size,
             array_in,
-            array_out,
+            grid_out.cl_array,
```
ondrejkrejci, line 418: ??

ondrejkrejci (commented):
Hi, having several questions on this.

ppafm/ml/Generator.py
NikoOinonen: This is about the GPU buffers. I don't think the KPFM has been implemented in the GPU version yet.

ppafm/ml/Generator.py
NikoOinonen: The change in the list format is because I wanted to separate the two different tip arrays into separate variables. This if-branch is the case when the tip density array is not set and point charges are used instead, so it just sets everything to `None`.

ppafm/ml/Generator.py
NikoOinonen: It is used when changing the FDBM parameters, specifically the exponent `B_pauli`. When the exponent changes, the array needs to be prepared again starting from the original array.

ppafm/ocl/AFMulator.py
NikoOinonen: The array is not lost. It still exists in the CPU memory. It's just deleted from the GPU memory here when the memory minimization is on.

ppafm/ocl/field.py
NikoOinonen: It was redundant. The array pointer is stored inside the `grid_out` object.

ppafm/ocl/field.py
NikoOinonen: No, I only implemented here for the GPU when I was testing different clamping types. I decided that the soft clamp is not really necessary, so I didn't bother putting it elsewhere. Do you think this would be useful as a function in the CPU code?

ppafm/ocl/field.py
NikoOinonen: Should be safe. The array preparation does not use the variable.

ppafm/ocl/field.py
NikoOinonen: Just eliminates one redundant variable.

ppafm/cli/gui/ppafm_gui.py
```diff
@@ -577,9 +609,14 @@ def loadInput(self, main_input, rho_sample=None, rho_tip=None, rho_tip_delta=Non
...
             self.bxV0.setDisabled(True)
             self.bxAlpha.setDisabled(True)
+            self.bxCutoffEnable.setDisabled(True)
+            self.bxCutoffEnable.setChecked(False)
+            self.bxCutoffValue.setDisabled(True)
         else:
             self.bxV0.setDisabled(False)
             self.bxAlpha.setDisabled(False)
+            self.bxCutoffEnable.setDisabled(False)
+            self.previous_cutoff = None
```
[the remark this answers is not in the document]
NikoOinonen: Yes, it's so that we know on each update if the cutoff has changed from the previous value. If it has changed then we apply the cutoff and redo the FFT on the tip density array. It's basically just to prevent redoing the same computation multiple times.

ppafm/ocl/field.py
NikoOinonen: Same as above.

ppafm/ocl/field.py
ondrejkrejci: Ok, this is something we should talk about in the meeting. As in principle, I like to have a lot of possibilities, there is already a lot of code unused, that is why.
Good for now,

ppafm/ocl/field.py
ondrejkrejci: Sorry, for my ignorance, what does `in_place` mean and why we need to set `array_out` when `in_place` is False?
Good to go, after this.

ondrejkrejci (approved):
Good to go from my point of view.

ProkopHapala:
I was recently using different `soft_clamp` function which is simpler (no need for exp) I think, but perhaps as long as it works it does not matter. 

I was trying to compare it with yours, but not sure if I use yours correctly.
https://colab.research.google.com/drive/1zM1QtqKDIx7ULy8geUq7FPVWitqfwK7R?usp=sharing

```python
import numpy as np
import matplotlib.pyplot as plt

def soft_clamp(y, dy, y1, y2):
    """
    Applies a soft clamp to y, smoothly transitioning values above y1 towards y2.
    Also computes the derivative dy accordingly using the chain rule.

    Parameters:
    - y: np.ndarray, input values to clamp
    - dy: np.ndarray, derivatives of y with respect to some variable x
    - y1: float, lower threshold for clamping
    - y2: float, upper threshold for clamping

    Returns:
    - y_new: np.ndarray, clamped y values
    - dy_new: np.ndarray, updated derivatives
    """
    y_new  = y.copy()
    dy_new = dy.copy()
    mask   = y_new > y1
    y12    = y2 - y1
    invdy  = 1.0 / y12
    z = (y_new[mask] - y1) * invdy
    y_new[mask]   = y1 + y12 * (1 - 1 / (1 + z))
    dy_new[mask] *= 1.0 / (1.0 + z)**2
    return y_new, dy_new

def soft_clamp_exp(array_in, min_value, max_value, width=10.0):
    """
    Soft clamp function using NumPy masks, smoothly clamping values.

    Parameters:
    - array_in: np.ndarray, input array
    - min_value: float, minimum clamp value
    - max_value: float, maximum clamp value
    - width: float, width of the transition region

    Returns:
    - array_out: np.ndarray, output array after applying the soft clamp
    """
    array_out = array_in.copy()

    # Mask for values greater than max_value
    if max_value < np.inf:
        mask_max = array_out > max_value
        array_out[mask_max] = (
            (array_out[mask_max] - max_value) / 
            (1 + np.exp((array_out[mask_max] - max_value) / width)) + max_value
        )
    
    # Mask for values less than min_value
    if min_value > -np.inf:
        mask_min = array_out < min_value
        array_out[mask_min] = (
            (array_out[mask_min] - min_value) / 
            (1 + np.exp((array_out[mask_min] - min_value) / -width)) + min_value
        )

    return array_out


# Example usage:
# y(x) = 1 / (x - 2)^6, derivative dy/dx = -6 / (x - 2)^7

# Define the domain, avoiding x = 2 to prevent division by zero
x  = np.linspace(0.0, 4.0, 1000)  # Start slightly above 2
y  = 1. / (x - 2.)**6
dy = -6. / (x - 2.)**7

y1 = 1.0
y2 = 3.0

# Apply the soft clamp
y_clamped, dy_clamped = soft_clamp(y, dy, y1, y2 )

y_clamped_exp = soft_clamp_exp(y, -y1, y1, width=(y2-y1) )

# Plotting the results
plt.figure(figsize=(12, 6))

# Plot y and its clamped version
plt.subplot(1, 2, 1)
plt.plot(x, y, label=r"$y(x) = \frac{1}{(x-2)^6}$", color='blue', alpha=0.5)
plt.plot(x, y_clamped, label="Clamped y", color='red')
plt.plot(x, y_clamped_exp, label="Clamped_exp  y", color='g')
plt.axhline(y1, ls='--', c='k', label="y1 and y2")
plt.axhline(y2, ls='--', c='k')
plt.title("Soft Clamping Function")
plt.xlabel("x")
plt.ylabel("y")
plt.legend()
plt.grid(True)
plt.ylim(0.0,5.0)

# Plot dy and its clamped derivative
plt.subplot(1, 2, 2)
plt.plot(x, dy, label=r"$dy/dx = -\frac{6}{(x-2)^7}$", color='blue', alpha=0.5)
plt.plot(x, dy_clamped, label="Clamped_exp dy/dx", color='red')
#plt.plot(x, dy_clamped_exp, label="Clamped_exp dy/dx", color='g')
plt.axhline(0, ls='--', c='k')  # Reference line for zero derivative
plt.title("Derivative of Soft Clamped Function")
plt.xlabel("x")
plt.ylabel("dy/dx")
plt.legend()
plt.grid(True)
plt.ylim(-15.0,15.0)

plt.tight_layout()
plt.show()
```


download

NikoOinonen:
We talked in the meeting that I should add the AIMS density to the pyridine example to demonstrate the problem and the fix. However, the artifacting that I observed in my tests does not really manifest itself noticeably with pyridine since the heaviest element there is nitrogen which does not have such high core density.

In my test I observed the effect at least with Cl-atoms, so some molecule with that or a heavier element would work better. Any suggestions which molecule we want? If not, then I will just pick something random.

@ondrejkrejci @mondracek @ProkopHapala

ProkopHapala:
> that or a heavier element would work better. Any suggestions which molecule we want? If not, then I will just pick something random.

What was quite nice you were testing GUI with some benzene substituted by -Cl, -Br groups if I remember. That could be quite nice. Maybe even more nice would be pyridine substituted by these groups halogen groups.

But in general pick whatever you you find convenient.

NikoOinonen:
> What was quite nice you were testing GUI with some benzene substituted by -Cl, -Br groups if I remember.

Yes, 1-bromo-3,5-dichlorobenzene. We actually already have it in the examples as an xyz file: https://github.com/Probe-Particle/ppafm/tree/main/examples/benzeneBrCl2.

> Maybe even more nice would be pyridine substituted by these groups halogen groups.

@ProkopHapala Something like this: https://pubchem.ncbi.nlm.nih.gov/compound/606256?

ProkopHapala:
Yes

mondracek:
@NikoOinonen did you test this PR in the CLI version? Did it work? I tried to run FDBI for some FHI-AIMS-generated XSF files and I keep hitting into the #312 issue. (I had to do some manual editing of the XSF files to overcome the problem that our code does not really understand these XSF from FHI-AIMS in their original form, but that does not seem to be the origin of the issue.)

NikoOinonen:
@mondracek I was just now trying to get it to work in the CLI, but I run into the exact same issue. @yakutovicha Do you think you can finish #316 soon?

NikoOinonen:
I added the example, but only for the GPU version until the CLI version issue gets fixed.

@ProkopHapala Can you add this file to the Zenodo record for the examples when you create it: https://www.dropbox.com/scl/fi/ilx7oe6iax375tqdleuax/BrClPyridine_hartree_density.tar.gz?rlkey=5j5xyc9n36zgw66xsb5mq8rvd&st=wyenjh4n&dl=0

NikoOinonen:
@mondracek Regarding using densities from aims, I just used the cube output and then converted it to xsf.

I used the following script for the conversion (not sure if the origin is handled correctly, so better to keep it at (0, 0, 0)):
```python

from pathlib import Path
from ase.io.cube import read_cube
from ppafm.io import saveXSF, primcoords2Xsf, Hartree2eV
import numpy as np

def cube_to_xsf(cube_path, xsf_path, scale=1.0):

    with open(cube_path, "r") as f:
        cube = read_cube(f, read_data=True)
        data = cube['data'].T * scale
        atoms = cube['atoms']
        origin = cube['origin']

    lattice = atoms.cell
    xyzs = atoms.get_positions() - origin
    Zs = atoms.get_atomic_numbers()
    lvec = np.concatenate([origin[None], lattice])

    xsf_path.parent.mkdir(exist_ok=True)
    atomstring = primcoords2Xsf(Zs, xyzs.T, lvec)
    saveXSF(xsf_path, data=data, lvec=lvec, head=atomstring)

if __name__ == "__main__":

    input_dir = Path(".")
    output_dir = Path(".")

    cube_to_xsf(input_dir / 'cube_001_hartree_potential.cube', output_dir / 'hartree.xsf', scale=Hartree2eV)
    cube_to_xsf(input_dir / 'cube_002_total_density.cube', output_dir / 'density.xsf', scale=1.0)
```

ondrejkrejci:
> @mondracek Regarding using densities from aims, I just used the cube output and then converted it to xsf.
> 
> I used the following script for the conversion (not sure if the origin is handled correctly, so better to keep it at (0, 0, 0)):
> 
> ```python
> from pathlib import Path
> from ase.io.cube import read_cube
> from ppafm.io import saveXSF, primcoords2Xsf, Hartree2eV
> import numpy as np
> 
> def cube_to_xsf(cube_path, xsf_path, scale=1.0):
> 
>     with open(cube_path, "r") as f:
>         cube = read_cube(f, read_data=True)
>         data = cube['data'].T * scale
>         atoms = cube['atoms']
>         origin = cube['origin']
> 
>     lattice = atoms.cell
>     xyzs = atoms.get_positions() - origin
>     Zs = atoms.get_atomic_numbers()
>     lvec = np.concatenate([origin[None], lattice])
> 
>     xsf_path.parent.mkdir(exist_ok=True)
>     atomstring = primcoords2Xsf(Zs, xyzs.T, lvec)
>     saveXSF(xsf_path, data=data, lvec=lvec, head=atomstring)
> 
> if __name__ == "__main__":
> 
>     input_dir = Path(".")
>     output_dir = Path(".")
> 
>     cube_to_xsf(input_dir / 'cube_001_hartree_potential.cube', output_dir / 'hartree.xsf', scale=Hartree2eV)
>     cube_to_xsf(input_dir / 'cube_002_total_density.cube', output_dir / 'density.xsf', scale=1.0)
> ```

This script should be in our [utilities](https://github.com/Probe-Particle/ppafm/tree/main/ppafm/cli/utilities).

NikoOinonen:
> This script should be in our [utilities](https://github.com/Probe-Particle/ppafm/tree/main/ppafm/cli/utilities).

I'm not sure about that. It's quite specific to this case when you are computing exactly the hartree potential and the density using aims. And again the origin is handled somehow weirdly: I had a cube with non-zero origin and after conversion it was displaying right in VESTA but was not working in ppafm. I think the real solution is to fix #190 so we don't have to make any conversions.

ondrejkrejci:
Well,

Just an idea - the origin is always wrong from AIMS, there is no way, so it would start from a real 0,0,0 - I am wondering, why it is not working with ppafm, but then it does not make sense.
Regards,
Ondrej

NikoOinonen:
@mondracek I added the CLI run script, so you can try it.

It's a little bit different from the GPU Python script, because it uses the dz2 tip for electrostatics instead of the delta density which had the wrong grid dimensions. I tried the same method for subtracting the valence electrons as in the pyridine example, but for some reason it was giving me really severe artefacts from the electrostatics so I did it this way. It's not ideal, but it demonstrates the effect of the cutoff.

NikoOinonen:
This is waiting for @mondracek to check that the CLI version runs correctly. Otherwise everything should be good, I think.

mondracek:
@NikoOinonen, I was able to reproduce the new FDBM feature with `ppafm/examples/BrClPyridine_FDBM_aims_density/`, but not if I have tried to generate `density.xsf` and `hartree.xsf` from my own `*.cube` files, even for the same BrCl-pyridine molecule. When I start from my own FHI-AIMS calculation, I am getting extreme artifacts in the FDBM images that are not influenced by the density cutoff option and the total electron density from my calculations seems to be an order of magnitude off anyway. Can you please send me the inputs for your DFT calculation, that is, your `geometry.in` and `control.in` files, best for both the sample molecule and the CO tip, and perhaps the version of the `fhi-aims` code you have used? Thanks.

NikoOinonen:
@mondracek Here are the aims input files and the script that I used for converting the cubes in to xsf:
[br-cl-pyridine-aims.zip](https://github.com/user-attachments/files/18688978/br-cl-pyridine-aims.zip)
Additionally, I had to manually round the lattice vectors (19.999... -> 20.0) in order to make it match exactly with the tip lattice:
```bash
sed -i -e 's/19.999935\(\s\|$\)/20.0 /g' -e 's/19.999935166934034\(\s\|$\)/20.0 /g' density.xsf
```
The molecule position and the grid position are chosen very specifically so that the center of the bromine atom lands exactly on a grid point in order to maximize the peak in the electron density.

>  the total electron density from my calculations seems to be an order of magnitude off

Sounds like could be an units issue. Notice in the conversion script that the Hartree potential is converted from Hartree to eV before saving to xsf, but the density is kept as is.

> geometry.in and control.in files, for [...] the CO tip

The CO tip is the same one used for the original pyridine example, I don't have the input files for it. I think it comes from VASP anyways.

mondracek:
Okay, thank you @NikoOinonen . So I finally get it. The problem was I was trying to use an AIMS-generated tip density and that seems to be so badly divergent that the `density-cutoff ` option does not help to repair it. This is unfortunate but I have no idea how to fix it at this moment. We may perhaps try to address the problem as a separate issue in the future. For now, I am giving green light to this PR.

mondracek (approved):

ProkopHapala:
> Okay, thank you @NikoOinonen . So I finally get it. The problem was I was trying to use an AIMS-generated tip density and that seems to be so badly divergent that the `density-cutoff ` option does not help to repair it. 

How it is possible ? 
What happens after we apply the cutoff ? 
I mean, if it is not enough we should perhaps cut it harder ;-) ...

NikoOinonen:
> The problem was I was trying to use an AIMS-generated tip density and that seems to be so badly divergent that the density-cutoff option does not help to repair it.

That seems strange, the cutoff should be applied to the tip as well. Did you try putting the cutoff lower, like 10, or even 1?

Anyways, thanks, I'll merge this now. I'll also make a new release so that the feature is available in the pip package.

PR #319, opened 2024-12-02
Add labels to GUI df curve plot.

### LINKED ISSUE #318: Calculated frequency change (df ) over labeled atomic sites as a function of tip-surface distance
Hi, when I click some specific atomic sites in ppafm-gui window, the curves came as below. So is them are the calculated force
spectroscopic measurements? Thanks!
df_curve

NikoOinonen:
Fixes #318 

Simply adds the x- and y-axis labels with the quantity and the unit to the df curve plots in the GUI:
df_curve


ondrejkrejci (approved):
I do not see any problems with this, feel free to go.

PR #331, opened 2025-01-31
Update upload-artifact v3->v4 in CI workflows.

NikoOinonen:
The CI workflow is broken as of yesterday, because [version v3 of `upload-artifact` is deprecated](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/). 

This updates the version to v4. Additionally, the name of the uploaded artifact in v4 cannot overlap between different jobs, so each python version uploads a different file.


yakutovicha (approved):
LGTM!

PR #335, opened 2025-02-10
Fix publish CI workflow

NikoOinonen:
The update to `upload-artifact` in #331 still left the publish workflow broken. This should fix it.

I was trying to make a new release, but this prevented it. I'll fix the tag to point to the fixed commit manually afterwards.


yakutovicha (approved):
go ahead!

PR #336, opened 2025-02-10
Add tutorials for the machine learning descriptors and data generator classes

### LINKED ISSUE #267: Tutorials for using the GPU-accelerated Python API
I want to add some tutorial type documentation for how to use the OpenCL Python API for running simulations, especially from the point of view of how to run a massive number of simulations like we have done for our machine learning training.

At least the following should be documented:
- [ ] The AFMulator
- [ ] The AuxMap descriptors
- [ ] The Generator that combines the previous two

NikoOinonen:
Fixes #267 

Adds some tutorials for how to use the Python API for generating the machine learning image descriptors and the data generator classes that wrap the AFM simulations and the descriptors into one interface.

A preview of the pages can be seen here: https://ppafm.readthedocs.io/en/generator-tutorial/tutorials/index.html


ondrejkrejci:
In https://ppafm.readthedocs.io/en/generator-tutorial/tutorials/generator-tutorial.html 

> distAbove which sets the tip-sample distance. The tip-sample distance here is measured as the distance between the top of the scan and the top atom in the sample subtracted by the vdW radius of the PP and the top atom of the sample.

Please define the way distance is defined and write an example, how should it be approximately set, at least for CO tip.

ondrejkrejci:
In https://ppafm.readthedocs.io/en/generator-tutorial/tutorials/afmulator-tutorial.html

> pixPerAngstrome sets the density of points in the force-field grid. The default value of 10 is usually adequate, but in some cases, a higher value can lead to slightly more accurate results, but at the expense of higher video memory usage. Try setting this lower if you get memory errors on devices with a low amount of video memory.

> scan_dim sets the number points in the x, y, and z directions in the scan region, and the scan_window sets the physical size of the scan region as a tuple with the start and end points (opposing corners) of the region in units of Ångströms.

I would suggest using in bold the force-field grid and scan region, to clearly see the difference, to what they manage.

ondrejkrejci:
> The amplitude of the oscillation is set by the df_steps parameter, which specifies how many steps in the z-direction are used in the conversion. The step size in z-direction is determined as (scan_window[1][2] - scan_window[0][2]) / scan_dim, so in this case the amplitude is 10 * (9.0Å - 5.0Å) / 40 = 1.0Å, and the final number of constant-height images is scan_dim[2] - df_steps + 1 = 40 - 30 + 1 = 31.
In future, let's have a new issue for using 2 possible parameters `df_steps` and `Amplitude` or something different, for better explanation. The `df_steps` an be than define as in cli like this: https://github.com/Probe-Particle/ppafm/blob/f5356e145a694919e41d1a4ae88a39be83111c99/ppafm/common.py#L382C5-L382C31

ondrejkrejci:
> The charge is set by specifying the magnitude (Qs) and the positions (QZs) of up to four point charges on the z-axis, with respect to the position of the probe particle. By using multiple charges it is possible to create multipole moments. In the above example we create a quadrupole charge distribution typical for a CO tip by placing three charges in a [q, -2q, q] pattern equidistantly. The quadrupole moment here comes to ÅÅ. For other tips, we may want to use a different charge distribution. For example, for a Xe tip a monopole charge is a better model:

I would just add there that the 4th item, is 0, but does not have any special meaning at the moment, the vectors are just `2*n`, because of OpenCL.

NikoOinonen:
@ondrejkrejci I think I addressed your comments. Could you have another look.

examples/Generator/inverse_trainer_xyz.py
```diff
@@ -72,6 +79,7 @@ def on_sample_start(self):
 
         # Iterate over samples in the batch
         for j in range(len(desc[0])):
+
```
ondrejkrejci, line 82: Why there is a single empty line, but it does not change anything.

NikoOinonen: I don't know, probably I just thought it's more aesthetically pleasing.

ondrejkrejci (approved):

ppafm/ml/AuxMap.py
```diff
@@ -41,6 +37,16 @@ def __call__(self, xyzqs=None, Zs=None, pot=None, rot=np.eye(3)):
             xyzqs[:, :3] = np.dot(xyzqs[:, :3] - xyz_center, rot.T) + xyz_center
         return self.eval(xyzqs, Zs, pot, rot)
 
+    @abstractmethod
```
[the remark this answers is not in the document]
NikoOinonen: Yes, the `@abstractmethod` means that the subclasses have to implement this method with this exact call signature. The idea is that we can be sure that the method exists and we can call any of the subclasses using the same input arguments.

PR #344, opened 2025-04-08
Enable saving tip displacements

### LINKED ISSUE #343: Saving tip displacement fails
There is an error message when trying to save tip displacements into `PPdisp_[xyz].xsf` files by calling `ppafm-relaxed-scan --disp`. The solution is trivial: The `perform_relaxation` function needs to be called with the `bPPdisp` argument set to `True` if the `--disp` option is set. I will submit the appropriate PR.

mondracek:
Enable `--disp` in `ppafm/cli/relaxed_scan.py`. Fixes #343


ondrejkrejci (approved):
It seems good to me. Have you checked that the tests/examples were running fine?

Issue #346, opened 2025-05-06
The option to rotate sample in the *xy*-plane does likely not work correctly


mondracek:
* There is an option `--rotate` available in `ppafm/cli/relaxed_scan.py` (or for the `ppafm-relaxed-scan` command). The option is described as "Rotates sampling in xy-plane". As I am working on #198, I was checking the order of indices for arrays that represent the force field components. The index order used to implement the rotation seems to be inconsistent with the index order elsewhere.  For example, the line
```ff_vdw[0, :, :, :], ff_vdw[1, :, :, :] = rotate_ff(ff_vdw[0, :, :, :], ff_vdw[1, :, :, :], opt_dict["rotate"])```
in `ppafm/cli/relaxed_scan.py` suggests that the first index distinguishes the vector components of the field, but the vector field has been loaded by the `io.load_vec_field` function a line before and this function, as far as I know, returns the data field in an arrangement in which the vector component is defined by the last index, not the first one.

* I tried to verify my suspicion that there is an error by running the `examples/Graphene` example with `ppafm-relaxed-scan -k 0.5 -q -0.05 --rotate 3.14159265` (which should correspond to 180 degree rotation) and comparing the resulting `examples/Graphene/Q-0.05K0.50/OutFz.xsf` file with the original file obtained without the `-rotate` option. The problem is, I am not sure I understand what exactly is the rotation supposed to do. However, I would expect something like the data in the `OutFz.xsf` file being flipped or rotated in some but otherwise the same. This is not really what I have seen when I looked at both versions of the `OutFz.xsf` with the Vesta viewer. So I think my suspicion is confirmed.

* The problem could likely be fixed just by flipping the array indices in those lines of `relaxed_scan.py` which call `rotate_ff`. I will test later whether that really helps. Also, I suspect nobody is actually using this feature, which is why the problem has not been notices earlier. Nevertheless, I have to admit I am rather happy for finding this bug, as it stresses the importance of doing what I propose in #198. Now it is not purely about making the code look nicer, there seems to be something really wrong in the present form of the code.

PR #347, opened 2025-06-02
Some changes in handling the arrays that represent force fields, potentials and densities

mondracek:
Partial step to address #198. Introduces the *xyz* order of numpy indices of arrays that represent fields in the CLI version. Unifies the functions that set the fields to be passed from a `numpy` array to the C++ code under one new function, `setFF` in `ppafm/HighLevel.py`.

### DIFF
--- a/ppafm/cli/plot_results.py
+++ b/ppafm/cli/plot_results.py
@@ -141,7 +141,6 @@ def main(argv=None):
                     dirname + "/xy" + atoms_str + cbar_str,
                     pp_positions[:, :, :, 0],
                     pp_positions[:, :, :, 1],
-                    slices=list(range(0, pp_positions.shape[2])),
                     BG=pp_positions[:, :, :, 2],
                     extent=extent,
                     atoms=atoms,
@@ -164,7 +163,6 @@ def main(argv=None):
                 PPPlot.plotImages(
                     dirname + "/IETS" + atoms_str + cbar_str,
                     iets,
-                    slices=list(range(0, iets.shape[2])),
                     zs=tip_positions_z,
                     extent=extent,
                     atoms=atoms,
@@ -177,7 +175,6 @@ def main(argv=None):
                 PPPlot.plotImages(
                     dirname + "/Evib" + atoms_str + cbar_str,
                     e_vib[:, :, :, 0],
-                    slices=list(range(0, iets.shape[2])),
                     zs=tip_positions_z,
                     extent=extent,
                     atoms=atoms,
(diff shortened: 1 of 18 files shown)


mondracek:
I forgot to mention that this PR also resolves #346.

ppafm/cli/plot_results.py
```diff
@@ -224,14 +219,14 @@ def main(argv=None):
             else:
                 # Prepare to calculate KPFM/LCPD
                 # For each pixel, find a,b,c such that df = aV^2 + bV + c
-                # This is done as polynomial (2nd order) regression, the above equality need not be exact
+                # This is done as polynomial (2nd order) regression, the above equality needs not be exact
```
NikoOinonen, line 222: It was actually grammatically correct before.

NikoOinonen (commented):
I didn't look at everything in detail, but I tried running all of the examples and this does seem to break the pyridineDensOverlap example. Some other examples were also broken but for other reasons: #348.

NikoOinonen:
@mondracek It specifically breaks the GPU version of the example. The code runs but the result is broken: 
df_004

mondracek:
@NikoOinonen, does the `pyridineDensOverlap` example work out correctly for you at least with the `main` branch of the code? In other words, are you sure the problem is specific to the changes made in the `cleanup` branch?
In any case, I would very much like testing it myself, but I still cannot make the OCL version make running on any of the machines available to me. You mentioned the OCL version can run on CPU too: Can you explain me how exactly can I achieve that (in Linux)? Also, I attach here the output and error message from `ppafm/examples/pyridineDensOverlap/run_gpu.py`: [output.txt](https://github.com/user-attachments/files/24109628/output.txt). Any idea, what can be the problem here? I ran this with the code in the `main` branch, so nothing related to my changes.

NikoOinonen:
> I have 2 GB of VRAM

This is not a lot, so that's probably the issue.

>  Is there a way to tell ppafm (or Python in general) which opencl-icd it should prefer?

Yes, on this line, change the `i_platform`: https://github.com/Probe-Particle/ppafm/blob/c0c8a9e46b65aac956afee9bc5da06aec58b7a2e/examples/pyridineDensOverlap/run_gpu.py#L14

You can get the platform numbers easily by running `ppafm-gui --list-devices`.

ProkopHapala:
> @NikoOinonen, I have 2 GB of VRAM. As for using POCL: I have already installed the `pocl-opencl-icd` package, but the `ppafm` code still seems to use GPU preferentially as long as it is available. Or so I believe, as it says `Initializing an OpenCL environment on NVIDIA CUDA` when the calculation begins. Is there a way to tell `ppafm` (or Python in general) which `opencl-icd` it should prefer?

I think it should be possible to choose specific OpenCL device by some command line argument but I do not remember which now, need to look in code or documentation

PR #371, opened 2025-11-04
Enable `edge` parameter in `plotImage`.

### LINKED ISSUE #370: Overlay atoms color optimization.
As the scanning region expands, the colors of the atoms are lost, leaving only black dots behind. I believe we are using black to represent the edges of the atoms. I am considering changing the edge color to match the fill color to avoid this issue, or alternatively, reducing the width of the edges.

Image

Image

huangchieh:
Fix #370 Now, the parameter `edge` is enabled. In some cases, when the scanning range is large, one can use `edge=False` to disable edge drawing. Then the correct color for the atoms reveal. 

```python
...
afmulator.plot_images(X=X, outdir=outfolder, prefix=f'df_atoms_depth{depth}', atoms=atoms_formatted, edge=False)
...
```

df_atoms_depth5_030

### DIFF
--- a/ppafm/PPPlot.py
+++ b/ppafm/PPPlot.py
@@ -40,11 +40,11 @@ def plotAtoms(atoms, atomSize=default_atom_size, edge=True, ec="k", color="w"):
         plt.fig.gca().add_artist(circle)
 
 
-def plotGeom(atoms=None, bonds=None, atomSize=default_atom_size):
+def plotGeom(atoms=None, bonds=None, atomSize=default_atom_size, edge=True):
     if (bonds is not None) and (atoms is not None):
         plotBonds(atoms, bonds)
     if atoms is not None:
-        plotAtoms(atoms, atomSize=atomSize)
+        plotAtoms(atoms, atomSize=atomSize, edge=edge)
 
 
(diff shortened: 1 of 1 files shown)


NikoOinonen:
> Ideally, the edge width could be adjusted automatically based on the cell size. To achieve this, we might need to pass the cell information to plotAtoms and scale both the edge width and atom size accordingly, so that atoms remain clearly visible for large unit cells.

I forget how matplotlib treats this, but it might be enough to just scale up the `figsize` in `plotImages`, if you could try that.

ondrejkrejci:
Hi @huangchieh and @NikoOinonen - Maybe there is a more elegant solution in MatPlotLib, but at least I personally am completely happy with just what is here, **but** it should be callable from outside - meaning - we should have it as parameter, that we can call (for now from params.ini later we will move to toml , if I remember correctly). Could you maybe find a way around it?

huangchieh:
The solution to #370 turned out to be unexpectedly simple. When calling the `plot_images()` method, I can specify `atomSize` directly, so nothing needs to be modified. I think we can reject and close this request. @NikoOinonen @ondrejkrejci

ondrejkrejci:
With this I am closing request.

PR #388, opened 2026-04-22
Fix broken tip displacement calculation

### LINKED ISSUE #387: Tip displacement output broken (--dist option)
I have broken the optional output of tip displacement (the `--dist` option to `ppafm-relaxed-scan` ) by my previous PR #378. The order of indices for the `init_pos` variable is confused [here](https://github.com/Probe-Particle/ppafm/blob/f11cbaace57317aa3d34ac9c48b4fdb9ffa4b9bc/ppafm/HighLevel.py#L176). I have a fix ready and I will publish it as soon as possible.

mondracek:
Fixes #387.

* The crucial line to be fixed was in `ppafm.HighLevel.perform_relaxation()`.
* I have added the `--dist` option to `examples/FFPB_KPFM/run.sh` so that we have some example that checks the `--dist` functionality in the future.
* The change I have made in the `relaxTipStrokes_omp` function in `ppafm/cpp/ProbeParticle.cpp` is optional. The old definition of `ioff` values combined with the new definition of index order resulted in an erratic scanning of tip position in the _x_,_y_ directions. The code still worked because all points would be eventually visited, although in a nonsensical order, but I thought it would still be better to make it consistent with the new index order.
* Finally, I have removed the `meshgrid3d` function from `ppafm/HighLevel.py`. It did nothing and could not do anything as it lacked any command for output: neither it had a return command nor did it any argument reassignment. It caught my attention when I was checking how the `np.meshgrid` function works in the line I had to fix.

### DIFF
--- a/examples/FFPB_KPFM/run.sh
+++ b/examples/FFPB_KPFM/run.sh
@@ -16,7 +16,7 @@ ppafm-generate-ljff -i LOCPOT_V0.xsf
 echo
 
 echo "Tip Relaxation"
-ppafm-relaxed-scan --Vrange -0.5 0.5 3
+ppafm-relaxed-scan --Vrange -0.5 0.5 3 --disp
 echo
 
 echo "Plotting"
--- a/ppafm/HighLevel.py
+++ b/ppafm/HighLevel.py
@@ -28,11 +28,6 @@ def symGauss(Evib, E0, w):
     return Gauss(Evib, E0, w) - Gauss(Evib, -E0, w)
 
 
-def meshgrid3d(xs, ys, zs):
-    Xs, Ys, Zs = np.zeros()
-    Xs, Ys = np.meshgrid(xs, ys)
-
-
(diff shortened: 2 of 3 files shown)


mondracek:
By the way, I had to commit this with `git commit --no-verify`, otherwise, I was getting
``` ERROR: Package 'isort' requires a different Python: 3.9.12 not in '>=3.10.0' ```.
As I understand it, it is because my  `git commit` calls an old version of Python for the pre-commit checks. Sorry for that, I do not have a full control over the software versions used on our computer grid. I suppose I could move my `ppafm` installation to some virtual environment where I would install my own `pip` and `git`, but it seemed to be too much hassle to do it at this moment (especially as I am not completely sure whether it would helped or not).

NikoOinonen (approved):
I see `ioff` was in the wrong order. I don't understand transpose order in the meshgrid, but I trust you tested that it works.

Issue #389, opened 2026-05-12
Adding atomSizes keyword to the ppafm-plot-results.py


ondrejkrejci:
In [plot_results.py](https://github.com/Probe-Particle/ppafm/blob/main/ppafm/cli/plot_results.py) there should be possibility to change the size of the atoms.

ondrejkrejci:
Closed by #390 .

PR #390, opened 2026-05-12
Add atomSize keyword to the ppafm-plot-results.py 

huangchieh:
To solve issue #389

- Add `atomSize` keyword to  ppafm-plot-results.py. 
- In the example of `PTCDA_single`, add some lines to demonstrate the usage. 
- To demonstrate this, `input_plot.xyz` is added, which is the same as `PTCDA.xyz`.

Default `atomSize=0.15`, as it was in `ppafm-plot-results.py`:
`ppafm-plot-results -k 0.5 -q -0.10 --arange 0.5 2.0 2 --df --atoms --atomSize 0.15`
df_atoms_000

`ppafm-plot-results -k 0.5 -q -0.10 --arange 0.5 2.0 2 --df --atoms --atomSize 0.5`
df_atoms_000


ondrejkrejci (commented):
Generally looks good to me, but I was wondering as @NikoOinonen did some time ago this class for all the important parameters, if we should add it there: https://github.com/Probe-Particle/ppafm/blob/a50afef51ce84cecfc1a629ab86c229b088782ca/ppafm/common.py#L26

NikoOinonen:
> Generally looks good to me, but I was wondering as @NikoOinonen did some time ago this class for all the important parameters, if we should add it there:
> 
> https://github.com/Probe-Particle/ppafm/blob/a50afef51ce84cecfc1a629ab86c229b088782ca/ppafm/common.py#L26

This wasn't my work, but it's probably not needed there since the plotting parameter is not passed to anywhere outside `plot_results`.

ondrejkrejci:
Thanks, then it is looking good to me.

ondrejkrejci (approved):

Issue #392, opened 2026-06-10
Adding a py3.14 into the build workflow


ondrejkrejci:
At the moment, we are supporting pythons 3.8 - 3.13 ; 3.14 is now used as well, so we should added into our automatic build workflow in https://github.com/Probe-Particle/ppafm/blob/a50afef51ce84cecfc1a629ab86c229b088782ca/.github/workflows/ci.yml#L10 .

ondrejkrejci:
Should we remove python 3.8 and 3.9 as it is end-of-life? https://devguide.python.org/versions/

NikoOinonen:
> Should we remove python 3.8 and 3.9 as it is end-of-life? https://devguide.python.org/versions/

It depends, do we want to have ppafm installable for those versions. If it still works, I don't see why we would remove it. We would also need to change the requirement here: https://github.com/Probe-Particle/ppafm/blob/a50afef51ce84cecfc1a629ab86c229b088782ca/pyproject.toml#L19

ondrejkrejci:
Ok, fair enough, I agree, but wanted to check this. Thanks!

ondrejkrejci:
This should be closed by #394

PR #394, opened 2026-06-11
adding 3.14 into pytests

ondrejkrejci:
Solves #392


NikoOinonen (approved):
Tests are passing, seems to be ok.

